PR #2, opened 2021-11-28
Fix furnace block fuel

sfence:
Similar problem like this https://github.com/minetest/minetest_game/pull/2895 is too in pipeworks.

By this, when the furnace uses the last fuel with replacement, a replacement can be lost or block fuel inventory. When it happens, no fuel can be inserted into the furnace and the blocking item cannot be taken out.

WITHOUT fix:
https://user-images.githubusercontent.com/17455197/143733434-507dab3e-1160-4aa2-b8a1-d46dbd68f96d.mp4

WITH fix:
https://user-images.githubusercontent.com/17455197/143733437-f5f71bfd-7d0b-4bd4-8707-d5c2efb935bb.mp4

PS: On pipeworks contentdb website, is still linked gitlab repository page, which doesn't exist now.

### DIFF
--- a/compat-furnaces.lua
+++ b/compat-furnaces.lua
@@ -189,8 +189,26 @@ local function furnace_node_timer(pos, elapsed)
 					fuel_totaltime = 0
 					src_time = 0
 				else
-					-- Take fuel from fuel list
-					inv:set_stack("fuel", 1, afterfuel.items[1])
+					print(afterfuel.items[1]:to_string())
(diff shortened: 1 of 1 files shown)


wsor4035:
> PS: On pipeworks contentdb website, is still linked gitlab repository page, which doesn't exist now.

we are aware(thank you for the reminder), that is the case with the other mods that have moved to mt-mods. sometime today hopefully me or buckaroo will change that

PR #6, opened 2022-01-02
Expose set_tube on public API

S-S-X:


OgelGames (approved):

S-S-X:
Should also add update_meta function.
Also having update_meta makes API more complete and set_tube less limited, set_tube will just change tube data and routing but wont update formspec.

BuckarooBanzay (approved):

PR #16, opened 2022-03-06
make pipeworks Game agnostic

wsor4035:
todo:
* [x] make it boot
* [x] make water pipes workoutside mtg
* [x] sounds
* [x] crafts
* [x] digging groups

direct support for the following games
* [x] mtg
* [x] farlands reloaded
* [x] mineclon*s
* [x] hades

related issues: https://github.com/mt-mods/technic/issues/214

maybe interesting to @sfence @BuckarooBanzay

### DIFF
--- a/autocrafter.lua
+++ b/autocrafter.lua
@@ -248,7 +248,7 @@ minetest.register_node("pipeworks:autocrafter", {
 	description = S("Autocrafter"),
 	drawtype = "normal",
 	tiles = {"pipeworks_autocrafter.png"},
-	groups = {snappy = 3, tubedevice = 1, tubedevice_receiver = 1},
+	groups = {snappy = 3, tubedevice = 1, tubedevice_receiver = 1, dig_generic = 1},
 	tube = {insert_object = function(pos, node, stack, direction)
 			local meta = minetest.get_meta(pos)
 			local inv = meta:get_inventory()
--- a/autocrafter.lua
+++ b/autocrafter.lua
@@ -424,12 +424,3 @@ minetest.register_node("pipeworks:autocrafter", {
 	},
 })
 pipeworks.ui_cat_tube_list[#pipeworks.ui_cat_tube_list+1] = "pipeworks:autocrafter"
-
-minetest.register_craft( {
-	output = "pipeworks:autocrafter 2",
-	recipe = {
-	        { "default:steel_ingot", "default:mese_crystal", "default:steel_ingot" },
-	        { "basic_materials:plastic_sheet", "default:steel_ingot", "basic_materials:plastic_sheet" },
-	        { "default:steel_ingot", "default:mese_crystal", "default:steel_ingot" }
-	},
-})
--- a/autocrafter.lua
+++ b/autocrafter.lua
@@ -179,10 +179,6 @@ local function update_meta(meta, enabled)
 			"image_button[3,2;1,0.6;pipeworks_button_" .. state .. ".png;" .. state .. ";;;false;pipeworks_button_interm.png]" ..
 			"list[context;src;0,4.5;8,3;]"..
 			"list[context;dst;4,0;4,3;]"..
-			default.gui_bg..
-			default.gui_bg_img..
-			default.gui_slots..
-			default.get_hotbar_bg(0,8) ..
 			"list[current_player;main;0,8;8,4;]" ..
 			"listring[current_player;main]"..
 			"listring[context;src]" ..
--- a/common.lua
+++ b/common.lua
@@ -183,7 +183,7 @@ function pipeworks.create_fake_player(def, is_dynamic)
 		is_player = delay(true),
 		is_fake_player = true,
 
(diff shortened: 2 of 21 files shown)


sfence:
I have https://github.com/sfence/pipeworks/tree/sfence_hades_revisited , so You can inspire by it, or I can add Hades support after code will be prepared for it.

sfence:
But for hades revisited support it should be kept in mind, that later, to not break main Hades ideas, it can be required to disable nodes like spigot, fountainhead.

wsor4035:
some of these recipes assume mesecons to be present, so I am unsure what the best way is to handle these crafts. I could just leave them for now?

sfence:
> some of these recipes assume mesecons to be present, so I am unsure what the best way is to handle these crafts. I could just leave them for now?
> 
> edit: deployer, dispenser, flow sensor, nodebreaker,

Messecons is part of MineClone game and is available for Hades Revisited too (same item names, different textures).
It looks like Mineclone uses same name for messecon wire too.

wsor4035:
ill just make everything but mesecons game agnostic, and a game can add aliases if they desire

S-S-X:
Noticed a lot of mixed spaces/tabs indentation in changed and added content.
Did not mention specific lines because it seems that most tables has inconsistent indentation.

Also in my opinion, if changing indentation, it is better to fix only changed/added and add another PR if changing formatting for everything, there's similar inconsistent indentation with old stuff.
If that's too hard then could as well ignore this here and do another PR just for formatting indentation.

autocrafter.lua
```diff
@@ -252,7 +248,7 @@ minetest.register_node("pipeworks:autocrafter", {
 	description = S("Autocrafter"),
 	drawtype = "normal",
 	tiles = {"pipeworks_autocrafter.png"},
-	groups = {snappy = 3, tubedevice = 1, tubedevice_receiver = 1},
+	groups = {snappy = 3, tubedevice = 1, tubedevice_receiver = 1, dig_generic = 1},
```
S-S-X, line 251: What is `dig_generic` group and where it is used / why it is added?

common.lua
```diff
@@ -183,7 +183,7 @@ function pipeworks.create_fake_player(def, is_dynamic)
 		is_player = delay(true),
 		is_fake_player = true,
 
-		_formspec = def.formspec or default.gui_survival_form,
+		_formspec = def.formspec or "size[8,8.5]label[1,1;404 formspec not found]",
```
S-S-X, line 186: This could probably be just empty string.
Got any idea why it is there in first place?

mod.conf
```diff
@@ -1,5 +1,6 @@
 name = pipeworks
 description = This mod uses mesh nodes and nodeboxes to supply a complete set of 3D pipes and tubes, along with devices that work with them.
-depends = default, basic_materials, screwdriver
-optional_depends = mesecons, mesecons_mvps, digilines, signs_lib, unified_inventory
+#depends = default, basic_materials, screwdriver
```
S-S-X, line 3:
```suggestion
```

trashcan.lua
```diff
@@ -25,10 +25,6 @@ minetest.register_node("pipeworks:trashcan", {
 				"item_image[0,0;1,1;pipeworks:trashcan]"..
 				"label[1,0;"..S("Trash Can").."]"..
 				"list[context;trash;3.5,1;1,1;]"..
-				default.gui_bg..
-				default.gui_bg_img..
-				default.gui_slots..
-				default.get_hotbar_bg(0,3) ..
```
S-S-X, line 31: Should these be conditionally included?
I've not idea how much this changes outcome but if it's useful somewhere then could be easily handled similar way how `pipeworks.liquids.water.source` is done.

autocrafter.lua
wsor4035: support for other games

trashcan.lua
wsor4035: no, because they return "" for legacy compat or are absolutely useless to the form

common.lua
wsor4035: probably because it trying to be a fake player as much as possible? dunno, dont think anyone ever cared to check it, so ill just set it to `or ""`

mod.conf
wsor4035: ah right. forgot to remove this, thanks

wsor4035: https://github.com/mt-mods/pipeworks/pull/16/commits/58bd68f79d3061f66b0df679c35ff6594ba56710 resolved

common.lua
wsor4035: https://github.com/mt-mods/pipeworks/pull/16/commits/58bd68f79d3061f66b0df679c35ff6594ba56710 resolved

S-S-X (approved):
Seems fine, did not test...
Indentation is still messed up but that's more "would be nice"-thing than "has to be resolved"-thing.

PR #26, opened 2022-05-06
Add option to forego real entities

TurkeyMcMac:
It seems like people sometimes have trouble with the performance of Pipeworks. Since the tube item entities are essentially decorative\*, I tried removing them. Doing this was quite simple, since Pipeworks can already handle luaentities without real entities bound to them. This PR simply gives the option of treating all mapblocks as inactive.

To test, I put 19 or so items into a loop of 9 segments. With real entities turned on, the average time per step taken by Pipeworks was about 920 microseconds, according to the profiler. With real entities turned off, the average time fell to about 660 microseconds. Without the entities, the server also will not need to send as many updates to the client, which was not measured by this benchmark.

When using real entities, it seems that a lot of the time is spent in `move_entities_globalstep_part1` updating the active blocks. Therefore, turning off real entities should give a performance boost even when no tube entities are spawned. Perhaps the checking of block activeness could be changed to use `minetest.compare_block_status` when available. The downside is that this would return true for forceloaded blocks, which could lead to more real entities being spawned if there are tubes in forceloaded blocks.

---

\* I believe TechPack's TubeLib is supposed to be a faster alternative to Pipeworks, and it goes without the entities.

### DIFF
--- a/default_settings.lua
+++ b/default_settings.lua
@@ -27,6 +27,7 @@ local settings = {
 	enable_cyclic_mode = true,
 	drop_on_routing_fail = false,
 	delete_item_on_clearobject = true,
+	use_real_entities = true,
 }
 
 pipeworks.toggles = {}
--- a/luaentity.lua
+++ b/luaentity.lua
@@ -73,31 +73,43 @@ local function get_blockpos(pos)
 	        z = math.floor(pos.z / 16)}
 end
 
-local active_blocks = {} -- These only contain active blocks near players (i.e., not forceloaded ones)
+local move_entities_globalstep_part1
+local is_active
 
-local move_entities_globalstep_part1 = function(dtime)
-	local active_block_range = tonumber(minetest.settings:get("active_block_range")) or 2
-	local new_active_blocks = {}
-	for _, player in ipairs(minetest.get_connected_players()) do
-		local blockpos = get_blockpos(player:get_pos())
-		local minp = vector.subtract(blockpos, active_block_range)
-		local maxp = vector.add(blockpos, active_block_range)
+if pipeworks.use_real_entities then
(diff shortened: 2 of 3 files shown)


BuckarooBanzay:
interesting feature :+1: 

side-note: i tested this some time ago and didn't really notice any performance gains but maybe i just did something wrong :shrug:

TurkeyMcMac:
Perhaps you left `move_entities_globalstep_part1` in place?

wsor4035:
works as intended. tested with a assortment of random tubes. did not verify any performance gain.

TurkeyMcMac:
Do you mean you haven't tested the performance, or did you test the performance and not find an improvement?

S-S-X (approved):
LGTM, did not test at all but might do some simple perf testing later.

S-S-X:
> Random thought, perhaps the opacity of the tubes could be turned down if this setting is enabled, as there is no real point to see in them anymore. Probably more work than it's worth

I think this is good idea but it probably would look way with some new / bit different style textures.
Not sure as I've not tried but might require bit more than just changing opacity.

Not that important for this PR anyway and could well be done through another PR if someone is interested.

wsor4035:
Random thought, perhaps the opacity of the tunes could be turned down if this setting is enabled, as there is no real point to see in them anymore. Probably more work than it's worth

wsor4035:
Github mobile is seriously broken. Anyways, changing the opacity can easily be done with texture modifiers, adding a separate set of textures is doubling the media aka loading time which isn't worth it

TurkeyMcMac:
By the way, here's a flamegraph I recorded with entities enabled using LuaJIT's profiler. This wasn't in a controlled environment, and I had some other mods active. I think it's useful nonetheless.

[prof.svg.gz](https://github.com/mt-mods/pipeworks/files/8641771/prof.svg.gz)

PR #28, opened 2022-05-08
Add option for opaque tubes

TurkeyMcMac:
This adds an option for giving tubes solid black backgrounds. It is meant to be enabled along with the option from #26. The textures don't look that good, but without custom opaque tube textures, this is to be expected.

### DIFF
--- a/decorative_tubes.lua
+++ b/decorative_tubes.lua
@@ -42,9 +42,6 @@ local pane_box = {
 	}
 }
 
-local texture_alpha_mode = minetest.features.use_texture_alpha_string_modes
-	and "clip" or true
-
 minetest.register_node("pipeworks:steel_pane_embedded_tube", {
 	drawtype = "nodebox",
 	description = S("Airtight panel embedded tube"),
@@ -55,7 +52,7 @@ minetest.register_node("pipeworks:steel_pane_embedded_tube", {
 		"pipeworks_pane_embedded_tube_sides.png",
 		"pipeworks_pane_embedded_tube_ends.png", "pipeworks_pane_embedded_tube_ends.png",
 		},
-	use_texture_alpha = texture_alpha_mode,
+	use_texture_alpha = pipeworks.tube_texture_alpha_mode,
 	node_box = pane_box,
 	selection_box = pane_box,
 	collision_box = pane_box,
--- a/default_settings.lua
+++ b/default_settings.lua
@@ -5,7 +5,6 @@ local prefix = "pipeworks_"
 local settings = {
 	enable_pipes = true,
 	enable_lowpoly = false,
-	enable_opaque_tubes = false,
 	enable_autocrafter = true,
 	enable_deployer = true,
 	enable_dispenser = true,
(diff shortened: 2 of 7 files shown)


S-S-X:
Should maybe set default based on entity setting?
* opaque tubes by default if entities disabled.
* transparent tubes by default if entities enabled.

Or should it even be selected based on only entity setting (no configuration at all)?
* opaque tubes if entities disabled.
* transparent tubes if entities enabled.

TurkeyMcMac:
I went with the first option you presented, in case people want transparent tubes with no entities.

wsor4035:
personally i think it would be best if this was linked to the no entities, and then only extended if someone really, really had a good reason for it to be separate.

TurkeyMcMac:
I can make that change if it's the consensus of the mod maintainers.

wsor4035:
i mean, is there a good reason to keep this as its own setting? if so keep it and ill approve it, else i would prefer the other way

TurkeyMcMac:
People might not like the look of the opaque tubes. That's the only reason I can think of.

wsor4035:
suggestion maybe? force two existing commits into one, and then add another for using non entity settings, and can be merged in as a no squash, allowing the one to be reverted later if people desire?

....or just do it and squash the pr on merge like normal

S-S-X:
Tested how it looks like, not worst thing I've seen but I think it would be better to update textures.

opaque1

I've tried simple thing applying `^[noalpha^[colorize:#BAA:80`:

opaque2

I think it would not be too hard to find some way to change textures making it look somewhat similar to original but making tubes opaque.
Did not test but some light silver/grayish could possibly work, possibly even grabbing pipe textures and using combine?
Maybe someone else could have better suggestions as I do not really have that much experience with textures.

If we are able to make textures look "good enough" maybe that could possibly be enough to drop configuration option and follow entity setting?

TurkeyMcMac:
How about this gray background for all tubes?
screenshot_20220508_174725

wsor4035:
could just use the default steel/iron block grey if you wanted to match. anyways, im no artist, so ill differ to anyone who is

SwissalpS:
might want to ping @El-Naso on this for artistic ideas.
Personally I prefer something not too dark, just opaque similar to the gray @TurkeyMcMac sugested last.

El-Naso:
I like how the lighter gray plays with the existing textures. Try d4d4d4? Also, will this work with custom textures? Mine, for instance, are 2px thinner which somehow doesn't create issues. Maybe I could make a simple background texture too.

El-Naso:
Try these, see if they look any better than plain gray. I'll tinker a bit with a few variations.

pipeworks_tube_bg_short
pipeworks_tube_bg_plain

S-S-X:
> Here are some tubes with background color #dadada.

I think this looks pretty good already and avoids having extra data and textures when transparent tubes are used.

Then it is possible to add conditional textures with `[png:<base64>`, option that could be considered for El-Naso's background textures.

wsor4035:
> It should work with custom textures. It should also be possible to override the background color/image in a texture pack. Here are some tubes with background color #dadada. screenshot_20220509_103224

this looks good. appears however the code is not pushed?

TurkeyMcMac:
I'll make a new PR.

TurkeyMcMac:
https://github.com/mt-mods/pipeworks/pull/29

PR #29, opened 2022-05-09
Make tubes opaque when entities are not used

TurkeyMcMac:
image


TurkeyMcMac:
I forgot to upload the texture, one moment.

wsor4035:
```lua
local function overlay_tube_texture(texture)
    -- It seems "pipeworks_opaque.png^(%s)" won't work due to the lack of texture scaling.
    return ("(%s)^pipeworks_opaque.png^(%s)"):format(texture, texture)
end

function pipeworks.make_tube_tile(tile)
    if pipeworks.use_real_entities then
        return tile
    elseif type(tile) == "string" then
        return overlay_tube_texture(tile)
    else
        tile = table.copy(tile)
        if tile.color then
            -- Won't work 100% of the time, but good enough.
            tile.name = tile.name .. "^[multiply:" .. minetest.colorspec_to_colorstring(tile.color)
            tile.color = nil
        end
        tile.name = overlay_tube_texture(tile.name)
        tile.backface_culling = nil -- The texture is opaque
        return tile
    end
end
```
no need to redefine the function twice within a if. additionally `due to the lack of texture scaling` `[resize` modifier exists, and no new media is needed for this pr as you can just do `[combine:16x16^[noalpha^[colorize:#ffffff` to make a texture of any solid color

TurkeyMcMac:
Texture packs might provide textures with different dimensions, no?

I added `pipeworks_opaque.png` so that texture packs can redefine the background color. There could be another way to do this. I don't know much about texture packs.

wsor4035:
> Texture packs might provide textures with different dimensions, no?
> 
> I added `pipeworks_opaque.png` so that texture packs can redefine the background color. There could be another way to do this. I don't know much about texture packs.

...if we where still living in the past (pre 5x at least) this would be true, but texture packs can do overrides, etc with there conf file

wsor4035:
https://github.com/minetest/minetest/blob/master/doc/texture_packs.txt#L156 read here down. using texture modifiers saves media being sent to the client, etc

S-S-X:
> using texture modifiers saves media being sent to the client(granted its one file, but one file here and there adds up)

For this it is good to also remember that modifiers will also be sent to client and AFAIK this is for each registered item instead of just once for every item using it.

Assuming this is true (I've not really checked but it seems to be true) and assuming that textures are fairly small:
* When modifiers are enabled they might easily add up sooner and become heavier than textures for both network and client.
* On the other hand when modifier are disabled they're zero cost for network and client while textures will add unnecessary overhead for both.

If it can be good enough with modifiers then I'd vote for those but also textures could be considered, if pushing for network performance then it would be good to also verify actual behavior.

TurkeyMcMac:
I removed the texture. I kept the double string substitution as I think it's necessary to deal with various resolutions from texture packs. Try [REFI Textures](https://content.minetest.net/packages/MysticTempest/refi_textures/) with Pipeworks for example.

wsor4035:
`When modifiers are enabled they might easily add up sooner and become heavier than textures for both network and client.` this is incorrect, unless your doing something like `[png`, then you *might* have a case

S-S-X:
> this is incorrect, unless your doing something like [png, then you might have a case

Like said I'm not sure if this is the case but it seems to be.

First take El-Naso's texture as 32x32 for example, it is 191 bytes, for small textures you'll probably stay under 1k.

Take some average texture with modifiers (from this pr), drop file name as it will be also for texture files, also do not consider any other modifiers as those are also for both cases, that leaves ~56 bytes.
Texture modifiers however if I've understood correctly are actually sent for every item, pipeworks seems to be registering around 300 tube items but only some are relevant here. Let's say 100 to make it simple, that would make 5600 bytes for modifiers raw.

I might be wrong about how it actually works as I've not checked but if it does what I think it does then this might also be too conservative not considering that texture string might actually be sent more than once per item.

Again, not sure if this is actually optimized in engine but I believe it is not really optimized for transfer other than possible compression which might or might not be relevant depending if it is specialized enough.

Do you know if this is wrong and engine wont actually do that?

wsor4035 (approved):

OgelGames:
What about broken tubes and panel embedded tubes?
image

TurkeyMcMac:
The broken tubes look fine IMO.

BuckarooBanzay (approved):

PR #32, opened 2022-05-14
Add configurable globalstep interval

TurkeyMcMac:
It is not always necessary to update tube entities every step, especially when real entities are not being used. This PR gives people the option of spreading out entity updates. The new code can handle precise intervals even with granular server steps, since the code dynamically adjusts to control the average interval. You can try it yourself by printing out `dtime_accum` every update.

I have found that the highest viable interval is 0.19. (With a server step of 0.09, this means 8/9 update intervals will be 0.18 and every ninth will be 0.27.) Above 0.19, accelerated entities move faster than the globalstep can manage.

Tube entities decelerate at different rates depending on the dtime, but I think that's a separate issue. I believe it could already manifest e.g. in singleplayer vs multiplayer.

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -32,8 +32,8 @@ end
 if pipeworks.toggles.pipe_mode == "pressure" then
 	minetest.log("warning", "pipeworks pressure logic mode comes with caveats and differences in behaviour, you have been warned!")
 end
-if pipeworks.globalstep_interval >= 0.2 and pipeworks.enable_accelerator_tube then
-	minetest.log("warning", "pipeworks accelerator tubes will not entirely work with a globalstep interval 0.2 or above.")
+if pipeworks.entity_update_interval >= 0.2 and pipeworks.enable_accelerator_tube then
+	minetest.log("warning", "pipeworks accelerator tubes will not entirely work with an entity update interval 0.2 or above.")
 end
 
 -- Random variables
--- a/init.lua
+++ b/init.lua
@@ -32,6 +32,9 @@ end
 if pipeworks.toggles.pipe_mode == "pressure" then
 	minetest.log("warning", "pipeworks pressure logic mode comes with caveats and differences in behaviour, you have been warned!")
 end
+if pipeworks.globalstep_interval >= 0.2 and pipeworks.enable_accelerator_tube then
+	minetest.log("warning", "pipeworks accelerator tubes will not entirely work with a globalstep interval 0.2 or above.")
+end
 
 -- Random variables
 
--- a/luaentity.lua
+++ b/luaentity.lua
@@ -383,7 +383,7 @@ local move_entities_globalstep_part2 = function(dtime)
 end
 
 -- dtime after which there is an update (or skip).
-local dtime_threshold = pipeworks.globalstep_interval
+local dtime_threshold = pipeworks.entity_update_interval
 -- Accumulated dtime since last update (or skip).
 local dtime_accum = 0
 -- Delayed dtime accumulated due to skipped updates.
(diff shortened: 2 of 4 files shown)


wsor4035:
i have tested the PR, it does work, im not sure of the merits of it. whats the use case for this? more performance? ability to send items faster? additionally, seems to be the sort of setting that when we get issues, person may say x happens, but you have to pry it out of them that oh, i cranked the value to something unrealistic. curious to see what others think.

wsor4035:
if this concept is approved, adding limits with log warnings would be ideal

TurkeyMcMac:
The purpose is performance. Items move at the same rate but with less frequent updates. I made the change since the globalstep can be a big performance cost.

wsor4035:
at what point do we add enough settings, that pipeworks turns into *the best performance is disabling the mod via settings*? anyways, i suppose this does make sense, as the engine has its own configurable step value. please add sane limits to this however and then i will approve it

TurkeyMcMac:
Yes, you could increase the server step interval, but that would affect all globalsteps. There might be some that need a higher granularity.

I made the maximum interval 1 because that's a viable interval when no accelerator tubes are used. Perhaps I could lower the maximum to 0.5 or so for safety. How about I add a warning like you suggested when the interval is 0.2 or more and accelerator tubes are enabled.

TurkeyMcMac:
Nevermind about the deceleration thing, I realized deceleration only applies when items move to the next tube segment anyway.

init.lua
```diff
@@ -32,6 +32,9 @@ end
 if pipeworks.toggles.pipe_mode == "pressure" then
 	minetest.log("warning", "pipeworks pressure logic mode comes with caveats and differences in behaviour, you have been warned!")
 end
+if pipeworks.globalstep_interval >= 0.2 and pipeworks.enable_accelerator_tube then
+	minetest.log("warning", "pipeworks accelerator tubes will not entirely work with a globalstep interval 0.2 or above.")
+end
```
fluxionary, line 37: what's this about?

luaentity.lua
```diff
@@ -382,13 +382,36 @@ local move_entities_globalstep_part2 = function(dtime)
...
 minetest.register_globalstep(function(dtime)
-	handle_active_blocks_timer = handle_active_blocks_timer + dtime
-	if dtime < 0.2 or handle_active_blocks_timer >= (dtime * 3) then
-		handle_active_blocks_timer = 0.1
-		move_entities_globalstep_part1(dtime)
-		move_entities_globalstep_part2(dtime)
+	if dtime >= 0.2 and dtime_delayed < 1 then
+		-- Reduce activity when the server is lagging.
+		skip_update = true
 	end
```
fluxionary, line 397: You can't assume dtime is normally less than 0.2. See e.g. https://github.com/minetest-monitoring/mesecons_debug for how to track whether things are laggy or not, and whether that's your fault or not.

TurkeyMcMac: I was just basing this part off the previous code. For all I know it isn't necessary.

TurkeyMcMac:
I did some speed tests on a server with a step of 0.09s. Here are the average costs of the relevant globalstep according to the profiler:

* Using real entities:
  * Interval of 0s: **950μs**
  * Interval of 0.19s: **715μs**
* Not using real entities:
  * Interval of 0s: **583μs**
  * Interval of 0.19s: **525μs**

BuckarooBanzay (approved):

fluxionary:
i feel like this update makes the performance of pipeworks even more inscrutable, but i haven't made the time to thoroughly review it. 

what we really need is an API that *any* mod can use to moderate its own behavior when the server is lagging, but that's a much larger scope than this current PR. basically, minetest needs a job scheduler, which should be part of the engine if possible. 

assuming this is about to get merged, i have one suggestion: don't use "globalstep interval" as a descriptor, use something like "pipeworks update interval". the globalstep is the globalstep, mods don't generally have their own.

BuckarooBanzay:
@fluxionary are you still having doubts about this? i'd like to merge this soon and don't see a problem with it...

fluxionary:
Good by me for now.

Issue #39, opened 2022-08-12
unusable `teleport_tubes` file after updating to `5.6.0`


BuckarooBanzay:
After updating pandorabox from `5.5.1` to `5.6.0` i got greeted with this:
```
minetest_1             | 2022-08-12 05:35:09: ERROR[Main]: /data/world//worldmods/pipeworks/teleport_tube.lua:48: attempt to index upvalue 'tp_tube_db' (a nil value)
minetest_1             | 2022-08-12 05:35:09: ERROR[Main]: stack traceback:
minetest_1             | 2022-08-12 05:35:09: ERROR[Main]: 	/data/world//worldmods/pipeworks/teleport_tube.lua:48: in function 'read_tube_db'
minetest_1             | 2022-08-12 05:35:09: ERROR[Main]: 	/data/world//worldmods/pipeworks/teleport_tube.lua:293: in function 'get_db'
```

Didn't investigate any further but it looks like the save-file was serialized with the new algo and somehow got corrupted :/

Here the files, before and after:
* [teleport_tubes_pandorabox_5.5.1.txt](https://github.com/mt-mods/pipeworks/files/9313933/teleport_tubes_pandorabox_5.5.1.txt)
* [teleport_tubes_pandorabox_5.6.0.txt](https://github.com/mt-mods/pipeworks/files/9313934/teleport_tubes_pandorabox_5.6.0.txt)

appgurueu:
1. This is way larger data than `serialize` should ever handle - consider splitting it across multiple files or storing it in a proper database;
2. `pipeworks` should wrap the call to `minetest.deserialize` in an `assert`;
3. Executing this with LuaJIT, I get

```
$ luajit teleport_tubes_pandorabox_5.6.0.txt 
luajit: teleport_tubes_pandorabox_5.6.0.txt:1: main function has more than 65536 constants
```

as expected; loading the 5.5.1 file works just fine however. The problem presumably is that the new serialization takes up one constant slot for every reference it uses (e.g. `r%X = ...`), halving the number of constants you can have.

The relevant issue is https://github.com/minetest/minetest/issues/7574

TurkeyMcMac:
Mod storage might work well, since the SQLite3 backend doesn't have to write everything when only one entry changes. There's some duplication of data in memory, but I have a PR open for that: https://github.com/minetest/minetest/pull/12562

TurkeyMcMac:
Here's a patch that I haven't tested at all:

```patch
[...]

PR #40, opened 2022-08-13
Store teleport tube DB in mod storage

TurkeyMcMac:
Advantages:
* Not limited by the maximum number of constants in `minetest.deserialize` (Fixes https://github.com/mt-mods/pipeworks/issues/39).
* Changes to individual entries can be committed without re-serializing the entire database, provided Minetest 5.5 or later is used.

If the mod storage is not initialized and there is an old `teleport_tubes` file present, the database is initially read from this file.

### DIFF
--- a/teleport_tube.lua
+++ b/teleport_tube.lua
@@ -1,8 +1,9 @@
 local S = minetest.get_translator("pipeworks")
-local filename=minetest.get_worldpath() .. "/teleport_tubes"
+local filename=minetest.get_worldpath() .. "/teleport_tubes" -- Only used for backward-compat
+local storage=minetest.get_mod_storage()
 
 local tp_tube_db = nil -- nil forces a read
-local tp_tube_db_version = 2.0
+local tp_tube_db_version = 3.0
 
 -- cached rceiver list: hash(pos) => {receivers}
 local cache = {}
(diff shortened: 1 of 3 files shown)


teleport_tube.lua
```diff
@@ -12,47 +13,81 @@ local function hash(pos)
...
+		end
+	else
 		local file_content = file:read("*all")
 		io.close(file)
 
+		local backup_filename = filename .. ".bak"
+		pipeworks.logger("Moving teleport tube DB to mod storage from " .. filename)
+		pipeworks.logger("Backing up old file as " .. backup_filename)
+		assert(os.rename(filename, backup_filename))
+
```
S-S-X, line 78: I'd say remove this file handling stuff completely and add note to readme about it.
Could wait for other opinions on this, that's just how I feel about it: no unnecessary stuff (also will it replace existing files or should there be check for that?)

TurkeyMcMac: I suppose that would work too. It's up to the maintainers.

SwissalpS: S-S-X makes a valid point: renaming means there would need to be one more check

TurkeyMcMac: Will do then.

TurkeyMcMac: Are you sure the message should be moved to the readme? IMO people are probably more likely to see it in the log.

wsor4035: > Are you sure the message should be moved to the readme? IMO people are probably more likely to see it in the log.

This, keep it in the log. As almost every other mod doing stuff like this sends it to the log/print/both. Any sys admin worth there salt is going to dig through a log to see what happened first, putting it in the Readme is unitutive

wsor4035: @TurkeyMcMac probably worth it to add a minetest.log line to the pipeworks.logger function, just for convenience

S-S-X: Keep logging, I mean note about old db file that can be removed. Logging is good but should happen only once when automatic upgrade happens while readme should tell where data is stored and tell what's changed.

TurkeyMcMac: @wsor4035 Do you mean that `pipeworks.logger` should just call `minetest.log`, or that it should call `minetest.log` in addition to `print`? I would favor the former.

wsor4035: i was thinking
```lua
pipeworks.logger = function(msg)
	  print("[pipeworks]: " .. msg)
	  minetest.log("action", "[pipeworks]: " .. msg)
end
```
however if you want to drop print, i wouldnt be opposed

S-S-X: I'd vote just log without print, depending on OS/config log can anyway end up to console and there print might just duplicate stuff printed to console.

wsor4035 (approved):
looks good, merging in 10 minutes (since others reviews have been resolved)

wsor4035:
bit more than 10 minutes since i got distracted, sorry

BuckarooBanzay:
nice work :+1:

Issue #24, opened 2022-04-24
wielder.lua:113: bad argument [#1] to 'ipairs' (table expected, got nil)


Bastrabun:
I do not have a repro unfortunately, but flux did a bit of research here: https://gitea.your-land.de/your-land/bugtracker/issues/1816

```
2022-04-23 10:31:57: ERROR[ConnectionSend]: con(4/1) dropped reliable packet for non existent peer_id: 899
2022-04-23 10:32:05: ERROR[Main]: ServerError: AsyncErr: Lua: Runtime error from mod 'xp_redo' in callback environment_Step(): ...ve/5.5.0/Minetest_live/bin/../mods/pipeworks/wielder.lua:113: bad argument [#1] to 'ipairs' (table expected, got nil)
2022-04-23 10:32:05: ERROR[Main]: stack traceback:
2022-04-23 10:32:05: ERROR[Main]: [C]: in function 'ipairs'
2022-04-23 10:32:05: ERROR[Main]: ...ve/5.5.0/Minetest_live/bin/../mods/pipeworks/wielder.lua:113: in function 'wielder_on'
2022-04-23 10:32:05: ERROR[Main]: ...ve/5.5.0/Minetest_live/bin/../mods/pipeworks/wielder.lua:147: in function 'action_on'
2022-04-23 10:32:05: ERROR[Main]: ...Minetest_live/bin/../mods/mesecons/mesecons/internal.lua:193: in function <...Minetest_live/bin/../mods/mesecons/mesecons/internal.lua:186>
2022-04-23 10:32:05: ERROR[Main]: ...etest_live/bin/../mods/mesecons/mesecons/actionqueue.lua:137: in function 'old_execute'
2022-04-23 10:32:05: ERROR[Main]: ...ve/worldmods/mesecons_debug/overrides/mesecons_queue.lua:20: in function 'execute'
2022-04-23 10:32:05: ERROR[Main]: ...etest_live/bin/../mods/mesecons/mesecons/actionqueue.lua:111: in function <...etest_live/bin/../mods/mesecons/mesecons/actionqueue.lua:73>
2022-04-23 10:32:05: ERROR[Main]: ...ive/5.5.0/Minetest_live/bin/../builtin/game/register.lua:425: in function <...ive/5.5.0/Minetest_live/bin/../builtin/game/register.lua:409>
```

For my reference 1816

wsor4035:
[...]

Issue #18, opened 2022-03-09
Add `pcall` around wielder calls


BuckarooBanzay:
Issues like this: https://github.com/mt-mods/signs_lib/issues/7 could be prevented simply by adding a `pcall` around
the deployer/node-breaker calls.

I'm aware this won't fix the root-problems but it would prevent some game-interrupting crashes.
The errors should also be logged and raised somewhere too.

Ideas/Thoughts?

S-S-X:
Very clear error messages with information about mod and function being called, stacktrace would be good.
This would be to allow finding mods that do not work well with pipeworks fake player.

I think this is important because pcall can also cause problems when mod code crashes, pcall just makes sure that server wont crash but it is still exception and might need special recovery.
Adding pcall without this information will essentially hide errors, it is important from beginning.

Yes, I do think that if enough debug information and clear errors are not added then it would be better to crash server.
Pcall is fine but good error logging is must, it is real error situation  and can lead to unexpected mod, player or world states.

OgelGames:
There used to be error handling in `monitoring` to prevent crashes, I'm not sure why it got removed, but I feel like there was probably some reason for it: https://github.com/minetest-monitoring/monitoring/commit/1a1bda6b4b37a21f3f398e02e760cd50b0e034ef

IMO the correct way to fix it would be to implement all the other functions into the fake player, possibly using the `__index` metamethod to cover unknown stuff.

S-S-X:
> IMO the correct way to fix it would be to implement all the other functions into the fake player

Problem with this is that most mods should not handle fake players but they attempt because mods do not take into account any other situation than real player.
Also remember that you'll also need to patch mintest core Lua API to make it "work" or problem stays.

I do think pcall is really only way to catch problem situations and make other mods handle situation correctly.
Some methods could be implemented but implementing complete API is not limited to player API but also includes all Minetest Lua API methods that expect player (userdata) instance as argument, sounds is one common for example.

Then player metadata API comes into play, there you'd have to make decision how to implement it. Should metadata be stored to real player account (mods can for example expect that some metadata is available if one field is present) or thrown away and always return... well, what?

fluxionary:
I've seen this issue crash something like a dozen mods at this point. Most recently, the trigger was a callback in petz when someone put food into a deployer. I was just about to come over here and propose both of the things Buckaroo already has. 

[...]

PR #42, opened 2022-09-16
check if target is protected before trying to break/place a node

fluxionary:
so as not to trigger protection violations. currently, there is a problem when

1. a player puts down a node breaker in an unprotected area
2. another player protects the area
3. nodes continue to be broken
4. protection violations occur, possibly w/ penalties on the violating player, who can no-longer even dismantle their machine.

### DIFF
--- a/wielder.lua
+++ b/wielder.lua
@@ -336,6 +336,14 @@ if pipeworks.enable_node_breaker then
 		masquerade_as_owner = true,
 		sneak = false,
 		act = function(virtplayer, pointed_thing)
+			local player_name = virtplayer:get_player_name()
+			if (
+				minetest.is_protected(pointed_thing.above, player_name) or
+				minetest.is_protected(pointed_thing.under, player_name)
+			) then
+				return
+			end
+
 			--local dname = "nodebreaker.act() "
 			local wieldstack = virtplayer:get_wielded_item()
 			local oldwieldstack = ItemStack(wieldstack)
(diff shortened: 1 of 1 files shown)


wsor4035 (approved):

S-S-X:
Note that this will also block some actions that can be done by actual player in protected area, probably not really important but it might affect some things.
I think better would have been to only test device position and let target specific checks handle target location (in that case violating player can still dismantle their machine).

OgelGames:
I don't see how this changes anything, the default `on_place` function already checks the protection, and any override to `on_place` should do the same.

SwissalpS:
> 4\. protection violations occur, possibly w/ penalties on the violating player, who can no-longer even dismantle their machine.

wouldn't the fix be for player A to protect his build in the first place? Why wait until another player has 'stolen' it.

fluxionary:
> I don't see how this changes anything, the default on_place function already checks the protection, and any override to on_place should do the same.

the default `on_place` checks the protection, the records a protection violation. the relevant server (your-land) uses a mod that tries to keep players from "glitching" through stuff by digging/placing nodes, so it levies increasing penalties on players who do so - eventually doing damage to the player after repeated attempts. a player was 20000 nodes away from the machine and kept getting killed. 

> wouldn't the fix be for player A to protect his build in the first place? Why wait until another player has 'stolen' it.

... that's right, blame the victim. the relevant details her are that player A was added to player B's place so they could build the machine, then was later removed from the protection area. 

> Note that this will also block some actions that can be done by actual player in protected area, probably not really important but it might affect some things.

possibly true. but i've already had to write an separate API to govern what can and can't go into the wielder machines to prevent crashes because the "fake player" doesn't support the game API the way other mods expect (#18, #24, etc.). so far as i'm concerned, the deployer/node breaker shouldn't be doing anything "weird" like eating food in the first place.

> I think better would have been to only test device position and let target specific checks handle target location (in that case violating player can still dismantle their machine).

i actually like this proposal better than the current solution, i'll redo this in another PR. not sure the relevant methods currently know where the machine is, but they probably could be informed.

S-S-X:
> possibly true. but i've already had to write an separate API to govern what can and can't go into the wielder machines to prevent crashes because the "fake player" doesn't support the game API the way other mods expect (https://github.com/mt-mods/pipeworks/issues/18, https://github.com/mt-mods/pipeworks/issues/24, etc.). so far as i'm concerned, the deployer/node breaker shouldn't be doing anything "weird", like eating food, in the first place.

Nothing weird necessarily, simplest example is machine that is activated with public mesecons button. Button is at protection edge, still in protected area. Button can be activated by anyone from unprotected area, it can be activated with deployer owned by another player who still cannot break button.

PR #43, opened 2022-09-17
check if machine is protected before trying to break/place a node

fluxionary:
this is slight modification to #42.

to recap:

so as not to trigger protection violations. currently, there is a problem when

1. a player puts down a node breaker in an unprotected area
2. another player protects the area
3. nodes continue to be broken
4. protection violations occur, possibly w/ penalties on the violating player, who can no-longer even dismantle their machine.

### DIFF
--- a/wielder.lua
+++ b/wielder.lua
@@ -336,6 +336,14 @@ if pipeworks.enable_node_breaker then
 		masquerade_as_owner = true,
 		sneak = false,
 		act = function(virtplayer, pointed_thing)
+			local player_name = virtplayer:get_player_name()
+			if (
+				minetest.is_protected(pointed_thing.above, player_name) or
+				minetest.is_protected(pointed_thing.under, player_name)
+			) then
+				return
+			end
+
 			--local dname = "nodebreaker.act() "
 			local wieldstack = virtplayer:get_wielded_item()
 			local oldwieldstack = ItemStack(wieldstack)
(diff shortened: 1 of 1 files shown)


SwissalpS:
this feels even weirder. Unfortunately I currently have no setup to test this out and understand why you are solving your problem this way :/

wsor4035:
please resolve conflicts @fluxionary

S-S-X:
> Unfortunately I currently have no setup to test this out and understand why you are solving your problem this way :/

Eye offset is needed because position of fake player when created is actually `position = vector.subtract(wielder_pos, assumed_eye_pos)` where `wielder_pos` is actual node position.
Similar might be possible using pointed_thing angle (a lot more complicated) but I think this is probably currently already simplest solution (unless new property for node pos is added but that wouldn't follow player api).

S-S-X (approved):
Have not tested but seems like it should do its job disallowing activation of deployer or node breaker if owner does not have access to said device.

wsor4035 (approved):

fluxionary:
@wsor4035: as there's clearly some qualms about my solution to this problem, i'd prefer that you not merge things so quickly in the future. i have no issue w/ arguing my position.

wsor4035:
@fluxionary an approval outside of my own...

fluxionary:
]> @fluxionary an approval outside of my own...

ah, so there is. but in general, it's totally fine to give these things at least a day or two before merging them, even w/ approval.

PR #45, opened 2022-09-20
Store teleport tube DB more compactly

TurkeyMcMac:
Entries are now stored in this format:

```
<can receive> ':' <channel>
```

The position is no longer stored in the value, since it is already hashed and used as the key.

With Minetest's current mod storage implementation, this change will reduce memory usage.

### DIFF
--- a/teleport_tube.lua
+++ b/teleport_tube.lua
@@ -3,7 +3,7 @@ local filename=minetest.get_worldpath() .. "/teleport_tubes" -- Only used for ba
 local storage=minetest.get_mod_storage()
 
 local tp_tube_db = nil -- nil forces a read
-local tp_tube_db_version = 3.0
+local tp_tube_db_version = 4
 
 -- cached rceiver list: hash(pos) => {receivers}
 local cache = {}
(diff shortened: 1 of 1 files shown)


wsor4035 (approved):
leaving this open for a bit since it is a new storage format

fluxionary:
> With Minetest's current mod storage implementation, this change will reduce memory usage.

i am curious, did you actually measure how much memory this saves? i'm not objecting to this change for any reason, or trying to trivialize it - it clearly reduces some redundancy, and that's a reasonable goal on its own account. but how many teleport tubes do you need for this to have a meaningful effect?

TurkeyMcMac:
Using a copy of the database from Pandorabox, approximately 2.3MB of RAM should be saved. That is, until my patch to Minetest gets accepted, probably before 5.7.

S-S-X:
Bit offtopic but kinda related

To reduce ram consumption significantly take a look at cache, when there was talk about implementing it looked like memory hog so I did some little testing: https://gist.github.com/S-S-X/4bcfcb9c4f433cb8bb15055fdd66d9ef
However better efficiency would require few other changes so simpler caching was implemented, here's also comparison using pandorabox's tp tube db back then:
```
Cache indexing by position:
21912.06640625 KB
18901 entries
Cache indexing by channel:
413.2578125 KB
5112 entries
```

fluxionary:
> Using a copy of the database from Pandorabox, approximately 2.3MB of RAM should be saved. That is, until my patch to Minetest gets accepted, probably before 5.7.

care to link the PR? i'm curious.

TurkeyMcMac:
> > Using a copy of the database from Pandorabox, approximately 2.3MB of RAM should be saved. That is, until my patch to Minetest gets accepted, probably before 5.7.
> 
> care to link the PR? i'm curious.

https://github.com/minetest/minetest/pull/12562

fluxionary:
what about the cost of just using mod_storage directly instead of caching things at all?

TurkeyMcMac:
That might be possible using some metatable magic to preserve backwards compatibility.

BuckarooBanzay (approved):

OgelGames:
Seems like a good optimization, but I'm not sure if the hash can be used as-is, there might be some data lost here: https://github.com/mt-mods/pipeworks/blob/b190d29d21ace3c283ca56547915af7d6430fc2c/teleport_tube.lua#L11-L13

SwissalpS:
> I'm not sure if the hash can be used as-is, there might be some data lost

if I understand correctly, hash_nod_position() concats the three axis into one long number.

Since tp-tube coordinates should be integers, that would leave 9 chars per axis (plus one for sign), with current MT worldsize limits, this should be fine.

Yes, should be tested to see if all those assumptions actually apply.

TurkeyMcMac:
The precision should be fine. `hash_node_position` uses 48 bits. I tried this in a Lua 5.1 REPL:

```lua
print(("%.30g"):format(2^48-1))
```

That prints the same thing as this:

```lua
print(("%.99g"):format(2^48-1))
```

PR #49, opened 2022-11-09
Add digilines to teleport tube and refactor code

### LINKED ISSUE #46: repairing teleport tubes w/ a hammer doesn't restore them to functionality.
a player reports:

When a sender teleport tube (tp) is broken, use hammer to repair. Now the tp tube sender is not working anymore. Expected: tp tube sender should tp the items as it did before breaking, doesn't work even after setting the channel name again. Workaround: destroy the receiver tube and then place it again and configure to get the sender tp tube working again.

### LINKED ISSUE #47: Cleanup tp tubes a bit
Remove:
https://github.com/mt-mods/pipeworks/blob/8c251800db1d9d190829cfdeea8cf339bcdecc89/teleport_tube.lua#L124-L136

Also remove these everywhere and just use direct `tp_tube_db`, just load db during mod loading stage instead. No real reason for deferred db read:
https://github.com/mt-mods/pipeworks/blob/8c251800db1d9d190829cfdeea8cf339bcdecc89/teleport_tube.lua#L114

Tp tube table `tp_tube_db` is local to file, recreating local for function scope in most cases also isn't helping with performance but instead just add unnecessary LoC.

OgelGames:
Adds the capability to set the teleport tube channel via digiline message, respecting private channels by adding an `owner` field on placement. Also refactors the code significantly to improve readability and maintainability.

Fixes #46 and closes #47

@BuckarooBanzay @S-S-X @TurkeyMcMac

### DIFF
--- a/teleport_tube.lua
+++ b/teleport_tube.lua
@@ -1,206 +1,304 @@
+
 local S = minetest.get_translator("pipeworks")
-local filename=minetest.get_worldpath() .. "/teleport_tubes" -- Only used for backward-compat
-local storage=minetest.get_mod_storage()
+local filename = minetest.get_worldpath().."/teleport_tubes"  -- Only used for backward-compat
+local storage = minetest.get_mod_storage()
 
-local tp_tube_db = nil -- nil forces a read
-local tp_tube_db_version = 4
+local has_digilines = minetest.get_modpath("digilines")
 
--- cached rceiver list: hash(pos) => {receivers}
-local cache = {}
+-- V1: Serialized text file indexed by vector position.
+-- V2: Serialized text file indexed by hash position.
+-- V3: Mod storage using serialized tables.
+-- V4: Mod storage using "<can_receive>:<channel>" format.
+local tube_db_version = 4
+local tube_db = {}
+local receiver_cache = {}
 
-local function hash(pos)
-	return string.format("%.30g", minetest.hash_node_position(pos))
+local function hash_pos(pos)
+	vector.round(pos)
+	return string.format("%.0f", minetest.hash_node_position(pos))
 end
 
--- New entry format: <can receive (number)> ':' <channel (string)>
-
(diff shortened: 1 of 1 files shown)


teleport_tube.lua
```diff
@@ -1,206 +1,304 @@
...
-	return {x = pos.x, y = pos.y, z = pos.z, cr = cr, channel = channel}
+local function deserialize_tube(hash, str)
+	local sep = str:sub(2, 2) == ":"
+	local cr = tonumber(str:sub(1, 1))
+	local channel = str:sub(3)
+	if sep and cr and channel then
+		local pos = minetest.get_position_from_hash(tonumber(hash))
+		return {x = pos.x, y = pos.y, z = pos.z, cr = cr, channel = channel}
+	end
 end
```
fluxionary, line 33: i'm not intimate w/ this code, but these don't seem equivalent at all. is "cr" *always* a single digit? why not do something like
```lua
local cr, channel = unpack(str:split(":", true, 1))
cr = tonumber(cr)
if cr and channel then
    local pos = minetest.get_position_from_hash(tonumber(hash))
    pos.cr = cr
    pos.channel = channel
    return pos
else
    return minetest.deserialize(str)
end
```

fluxionary: oh i see it's "can receive" and is either 0 or 1? so i guess it's always a single number.

fluxionary: and the de-serialize stuff is handled in the update code. got it.

OgelGames: Yes, `cr` (short for `can_receive`) is always either 0 or 1, so it's much more efficient to use `string.sub`.

wsor4035 (approved):
did not test it(can if you wish), but reading over the code, it makes sense

BuckarooBanzay (approved):

PR #51, opened 2022-11-10
Prevent vacuum tubes breaking themselves and refactor code

### LINKED ISSUE #48: Slow down vacuum tubes so they don't break themselves
I.e. only vacuum a limited number of stacks per second, instead of trying to pick up everything at once.

OgelGames:
Limits the items picked up per cycle to half of `pipeworks_max_items_per_tube`. Also refactors the code with some minor optimizations, including switching to node timers instead of an ABM.

Fixes #48

@BuckarooBanzay @S-S-X

### DIFF
--- a/vacuum_tubes.lua
+++ b/vacuum_tubes.lua
@@ -1,4 +1,49 @@
+
 local S = minetest.get_translator("pipeworks")
+
+local enable_max = minetest.settings:get_bool("pipeworks_enable_items_per_tube_limit", true)
+local max_items = tonumber(minetest.settings:get("pipeworks_max_items_per_tube")) or 30
+max_items = math.ceil(max_items / 2)  -- Limit vacuuming to half the max limit
+
+local sqrtof3 = math.sqrt(3)  -- Distance between opposite corners of a 1x1x1 cube
+
+local function inside_area(p, minp, maxp)
+	if p.x < minp.x or p.x > maxp.x then return false end
+	if p.y < minp.y or p.y > maxp.y then return false end
+	if p.z < minp.z or p.z > maxp.z then return false end
+	return true
+end
+
+local function vacuum(pos, radius)
+	radius = radius + 0.5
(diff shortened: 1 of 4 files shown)


vacuum_tubes.lua
```diff
@@ -1,107 +1,114 @@
...
+	if p.z < minp.z or p.z > maxp.z then return false end
+	return true
+end
+
+local function vacuum(pos, radius)
+	radius = radius + 0.5
+	local min_pos = vector.subtract(pos, radius)
+	local max_pos = vector.add(pos, radius)
+	local count = 0
+	for _, obj in pairs(minetest.get_objects_inside_radius(pos, radius * sqrtof3)) do
```
fluxionary, line 22: why not use `minetest.get_objects_in_area`? you could dump the `sqrtof3` and `inside_area` stuff.

https://github.com/minetest/minetest/blob/6191bafcadc21277be5527ed1ac05f7902e710ad/doc/lua_api.txt#L5509-L5511

OgelGames: Hmm... I forgot about that one 🤔 The `min_minetest_version` for pipeworks is 5.2.0, and that function was added in 5.4.0, so I'll have to add a check for that function.

S-S-X: Or maybe just bump min version to 5.4, those with old engine can keep using previous release.
Allows using new API with future improvements and cleanup stuff.

OgelGames: Yeah, I was thinking the same thing, it will make things easier moving forward 👍

wsor4035: bumping to 5.4 makes sense as some distros are still on that version, and multicraft is minetest 5.4 based.

BuckarooBanzay (approved):

BuckarooBanzay:
haven't tested but LGTM :+1:

Issue #17, opened 2022-03-06
make all formspecs use minetest.show_formspec rather than legacy formspec meta mess


wsor4035:
what the title says, for obvious reasons, like upgradability, etc

OgelGames:
Since when are metadata formspecs legacy? IMO they are better for simple formspecs.

wsor4035:
no, they really are not better, as it typically requires a lbm to update them as they are hardcoded into meta.

wsor4035:
for reference https://github.com/mt-mods/travelnet/issues/39

S-S-X:
There's reasons to drop metadata semi-static formspecs but if there's no real reason then dropping metadata formspecs is downgrade instead of upgrade.
Fully dynamic formspecs are a lot worse for performance while metadata formspecs are cached client side immediately during mapblock transfer.

OgelGames:
> for reference [mt-mods/travelnet#39](https://github.com/mt-mods/travelnet/issues/39)

I completely agree with the use there, travelnet formspecs are big and, more significantly, are different depending on who uses them. Most of the pipeworks formspecs are small and rarely need updating.

sirrobzeroone:
Oddly I'm here because I'm working on wine mod which has a pipe interface but stumbled on this issue. I've handled this in the wine mod updates I'm doing, I wouldn't really class formspec stored in the meta of a node as hard coded. This can be checked and updated on every right-click if you need to. Sure it's a little more work but you can set a version flag if you need to so you don't update the meta every time. It does mean you need code in the on rightclick event, most node formspecs are normally stored somewhere as a template in the mod code. Not sure if pipeworks uses a timer but if it does that's already calling every X seconds and can also be used as a hook for formspec refreshes - the furnace from default for example updates its node formspec every second when it's active :) 

Here is some quasi-code as an example, sorry I'm adapting this on the fly so it may not be 100% correct:

inside nodes on_rightclick 
```		-- Convert Check for node meta fs change
		local meta = minetest.get_meta(pos);
		if meta:get_int("version") <= 1 then
			mod_name.valid_formspec(meta) 		
		end
```

Inside mod_name.valid_formspec(meta)
```
		-- Could be as simple as
		meta:set_string("formspec", mod_name.formspec(pos))

		-- or you can get complex and go
		local new_meta = meta:set_int("new_meta", 1)

		local node_inv = meta:get_inventory()
		node_inv:set_size("new_inv", 1)

[...]

PR #57, opened 2023-01-23
Use sane default i3 inventory settings when missing

Awkanimus:
- Both `i3.settings.hotbar_len` and `i3.settings.inv_size` were removed in `i3`'s commit `75fddf7` because they are now per-player settings in `I3`
 - This just guesses sane defaults when the settings are not present to avoid a few nils and allow the game to continue without addressing the fact that these settings are now per-player. If there is a mixture of player settings (ie some with inventory sizes of 9, and some with 8), oddness will ensue despite the lack of crash.
- Note: Left in the removed settings for now to make this drop-in safe ; they should be removed at some point.

Tested in my soup where `i3.settings.legacy_inventory` defaults to true ; The Stackwise Filter-Injector has an inventory 8-wide.

Given the referenced commit above (search for `hotbar_len`) and the changes here, I think I've done enough testing; let me know if more is desired. I've run out of time for now.

Luacheck tells me it's all clean.

EDIT: Punctuation and spacing

### DIFF
--- a/common.lua
+++ b/common.lua
@@ -149,7 +149,8 @@ function fs_helpers.get_inv(y)
 		local inv_x = i3.settings.legacy_inventory and 0.75 or 0.22
 		local inv_y = (y + 0.4) or 6.9
 		local size, spacing = 1, 0.1
-		local hotbar_len = i3.settings.hotbar_len
+		local hotbar_len = i3.settings.hotbar_len or (i3.settings.legacy_inventory and 8 or 9)
+		local inv_size = i3.settings.inv_size or (hotbar_len * 4)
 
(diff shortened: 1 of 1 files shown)


wsor4035:
related https://github.com/mt-mods/pipeworks/issues/17

BuckarooBanzay (approved):

BuckarooBanzay:
@wsor4035 anything blocking this? look trivial IMO

S-S-X (approved):
LGTM, like Buckaroo said looks like pretty simple compatibility update.

wsor4035:
> @wsor4035 anything blocking this? look trivial IMO

doesnt properly fix it(not per player as it should be), but at least makes it work which is good enough for now

PR #66, opened 2023-05-16
log when items teleport through tubes

fluxionary:
i'm trying to figure out where a player's items went after they were passed through a teleport tube, but didn't arrive at the destination. this PR adds a simple log to aid in resolving future issues.

### DIFF
--- a/teleport_tube.lua
+++ b/teleport_tube.lua
@@ -233,7 +233,8 @@ local function can_go(pos, node, velocity, stack)
 	velocity.x = 0
 	velocity.y = 0
 	velocity.z = 0
-	local channel = minetest.get_meta(pos):get_string("channel")
+	local src_meta = minetest.get_meta(pos)
+	local channel = src_meta:get_string("channel")
 	if channel == "" then
 		return {}
 	end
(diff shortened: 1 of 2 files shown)


S-S-X:
I believe this should be configurable and disabled by default. It can produce a lot of unnecessary logging when there's hundreds of stacks going through system.

But I've also noticed items kind of disappearing, not really disappearing but getting stuck in teleport tube loop. So far all the stuff seemed to be returning when loop is broken, tubetool been helpful as it can list connected tubes, display their config and reveal locations.

S-S-X:
By a lot I mean significant portion of these are going through tp tubes all the time:
image

fluxionary:
is that 1.5k entities going through teleport tubes *per second*? that certainly warrants a setting, will do

teleport_tube.lua
```diff
@@ -242,6 +243,14 @@ local function can_go(pos, node, velocity, stack)
 		return {}
 	end
 	local target = receivers[math.random(1, #receivers)]
+	local src_owner = src_meta:get_string("owner")
```
BuckarooBanzay, line 246: not sure if this is performance-critical, but can this be moved inside the `if` clause below? thanks :+1:

fluxionary: admittedly sloppy on my part, i've rescoped those 3 calculations.

S-S-X (approved):

PR #68, opened 2023-05-26
Full digilines supporting tp tubes

SwissalpS:
Have done some shallow testing.
I'm open for discussion and suggestions, it was kind of either try and fix tube-tool or add better digilines support to tp-tubes :D

### DIFF
--- a/teleport_tube.lua
+++ b/teleport_tube.lua
@@ -259,14 +259,32 @@ local function repair_tube(pos, node)
 	update_meta(meta)
 end
 
-local function digiline_action(pos, _, channel, msg)
+local function digiline_action(pos, _, digiline_channel, msg)
 	local meta = minetest.get_meta(pos)
-	if channel ~= meta:get_string("digiline_channel") or type(msg) ~= "string" then
+	if digiline_channel ~= meta:get_string("digiline_channel") then
 		return
 	end
-	local name = meta:get_string("owner")
-	local cr = meta:get_int("can_receive")
-	update_tube(pos, msg, cr, name)
+	local channel = meta:get_string("channel")
+	local can_receive = meta:get_int("can_receive")
+	local t_msg = type(msg)
+	if t_msg == "string" then
+		channel = msg
+	elseif t_msg == "table" then
+		if type(msg.channel) == "string" then
+			channel = msg.channel
+		end
+		if nil ~= msg.can_receive then
+			if msg.can_receive == 1 or msg.can_receive == true then
+				can_receive = 1
+			else
+				can_receive = 0
+			end
+		end
+	else
+		return
+	end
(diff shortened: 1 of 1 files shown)


S-S-X:
Not sure if this is off topic or not but do you mean tube tool still does not work after attempted fix switching to new API?

SwissalpS:
> Not sure if this is off topic or not but do you mean tube tool still does not work after attempted fix switching to new API?

(Yeah, soff-topic)  
I've checked out at b75978c09 and it didn't copy 'can_receive' flag

teleport_tube.lua
```diff
@@ -270,14 +270,32 @@ local function repair_tube(pos, node)
...
+		if nil ~= msg.can_receive then
+			if msg.can_receive == 1 or msg.can_receive == true then
+				can_receive = 1
+			else
+				can_receive = 0
+			end
+		end
+	else
+		return
+	end
```
OgelGames, line 296:
```suggestion
	if type(msg) == "string" then
		channel = msg
	elseif type(msg) == "table" then
		if type(msg.channel) == "string" then
			channel = msg.channel
		end
		if msg.can_receive == 1 or msg.can_receive == true then
			can_receive = 1
		elseif msg.can_receive == 0 or msg.can_receive == false then
			can_receive = 0
		end
	else
		return
	end
```

OgelGames (changes requested):

teleport_tube.lua
SwissalpS: why prefer calling type() multiple times from assigning a var?  
is the reason for readability or for performance?

S-S-X: Probably no real difference with performance but IMO calling type grants better readability and is easier to follow.

For performance: introducing new local variable also has little performance impact, it is not instant operation. Static strings are very well optimized and type should also be very cheap operation.
I wouldn't be surprised if local variable would actually be slower but it does not really matter, in actual game difference probably cannot be even measured.

S-S-X (approved):
LGTM
Personally I do not like being able to directly program everything in game but can't really come up with any easy interesting ways to do it in this case.

> I've checked out at b75978c09 and it didn't copy 'can_receive' flag

Checked this one, it was actually working but failed to update formspec. If you've used it and wondered why tp tubes do weird things this might be reason (config works but form show previous receive setting). Fixed formspec update issue already so should work with latest version.

wsor4035 (approved):
seems fine

teleport_tube.lua
OgelGames: Yeah, `type()` is such a small function that calling it repeatedly is better for readability than using a variable. And like @S-S-X said, there's practically no performance difference.

OgelGames (approved):

PR #69, opened 2023-06-11
Fix autocrafter destroying return items that don't fit in dst

### LINKED ISSUE #61: autocrafter: if there isn't enough room in the output inventory, replacements may be lost

image

image

SwissalpS:
see https://github.com/mt-mods/pipeworks/issues/61

fixes #61

also snuck in two other minor changes


SwissalpS:
compare with #70 and decide which is better

BuckarooBanzay:
haven't tested it but this looks cleaner :+1:

S-S-X (approved):
Didn't test but looks fine.

SwissalpS:
closing to give #70 preference, can be reopened if there is reason to.

PR #70, opened 2023-06-11
Fix autocrafter variant 2

### LINKED ISSUE #61: autocrafter: if there isn't enough room in the output inventory, replacements may be lost

SwissalpS:
see #61 and #69

alternative to #69

fixes #61

### DIFF
--- a/autocrafter.lua
+++ b/autocrafter.lua
@@ -68,6 +68,8 @@ local function autocraft(inventory, craft)
 
 	-- success, so apply to actual output inventory
 	inventory:set_list("dst", sandbox_inv:get_list(""))
+	-- clear sandbox, just in case it matters
+	sandbox_inv:set_list("", {})
 	-- consume materials
 	for itemname, number in pairs(consumption) do
 		for _ = 1, number do -- We have to do that since remove_item does not work if count > stack_max
--- a/autocrafter.lua
+++ b/autocrafter.lua
@@ -97,18 +97,18 @@ end
 -- note, that this function assumes allready being updated to virtual items
 -- and doesn't handle recipes with stacksizes > 1
 local function after_recipe_change(pos, inventory)
+	local hash = minetest.hash_node_position(pos)
 	local meta = minetest.get_meta(pos)
 	-- if we emptied the grid, there's no point in keeping it running or cached
 	if inventory:is_empty("recipe") then
 		minetest.get_node_timer(pos):stop()
-		autocrafterCache[minetest.hash_node_position(pos)] = nil
+		autocrafterCache[hash] = nil
 		meta:set_string("infotext", S("unconfigured Autocrafter"))
 		inventory:set_stack("output", 1, "")
 		return
 	end
-	local recipe = inventory:get_list("recipe")
 
-	local hash = minetest.hash_node_position(pos)
+	local recipe = inventory:get_list("recipe")
 	local craft = autocrafterCache[hash]
 
 	if craft then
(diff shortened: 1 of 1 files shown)


S-S-X:
If creating temporary inventory it should be cleared always after use, never leave stacks in temporary node inventories.

S-S-X:
I don't really like global temporary inventory concept utilized here, it just feels really wrong. It probably works but is potential source for future issues because constraints are harder to understand compared to localized variant.

SwissalpS:
> I guess it is not possible to take items from detached inventory without explicit server side code?

I don't think so. Only ways I know are formspecs and code for any type of inventory.

SwissalpS:
> I don't really like global temporary inventory concept utilized here

Both solutions #69 and #70 are a bit hacky, yes. Do you think it would be better for each autocrafter to have a shadow inventory in it's meta?

Wouldn't that, even if they were empty most of the time, add quite a bit of overhead?

SwissalpS:
searching through pandorabox-mods, I find about 10 instances using detached inventories.
The most similar is unified_inventory_plus that uses it to test crafting very much like this branch does.

Differences: it passes playername and destroys it right after every use.

I do encourage everyone to test this in what ever way possible :D

SwissalpS:
playername argument will be removed in future, so let's not use it now.

If this last commit makes overhead or danger for cruft buildup any less or more likely, I cannot tell. What it does is gives sandbox inv less lifetime and confined to method scope.

S-S-X:
> Wouldn't that, even if they were empty most of the time, add quite a bit of overhead?

No idea, I guess it mostly depends on storage. If detached inventory is written to storage then both should be about same, both will get exact same number of updates and writes. If detached inventory wont get automatically written to storage then sure difference is significant.

Writing mapblock storage vs writing some other storage should be same. If both are written to storage then both probably are also utilizing deferred writes (afaik mapblocks do and first stage is backed by ram) which in the end shouldn't be writing anything if lists are always cleared after use.

But afaik mt engine simply marks things dirty instead of actually tracking changes so it probably still always attempts to make sure that data isn't there. Overhead however shouldn't really be significant thing because metadata for position will be anyway written no matter what.

That alternative approach modifies original list so no storage overhead but it does have a little bit "syscall" overhead because of loop, which one is actually faster heavily depends on what engine does. Maybe just run some benchmarks with both.

> Not using a given technology just because we don't understand it, does not seem a good reason not to use it and learn to understand and apreciate it.

Reason why I do not like is not related to understanding how detached inventory works, it is related to understanding how single global storage works. Basically I just do not like idea of having something global when it seems that local is more appropriate.

But I do also know it might be simplest way to get it through because best way to do this would be either implementing specialized method to check inventory space or implementing pure Lua inventory lists as a library and using that.

> Differences: it passes playername and destroys it right after every use.

Player name does not really make any sense here because it is bound to position instead of player, destroying list after every use anyway sounds about right just because it is not needed after or between uses. Destroying list however probably wont really do much good here, maybe keeps it from saving a little bit of unnecessary data if detached lists are backed by some storage engine.

SwissalpS:
Thanks for elaborating :D
It seems detached inventories are not stored (or were not) as mods like armor reinit the detached inv whenever player joins and fill it from player meta.

fluxionary:
every action on a detached inventory is sent to all players. this will potentially generate a huge quantity of spam packets. the solution is to instead create a "fake inventory" e.g. https://github.com/fluxionary/minetest-futil/blob/main/minetest/fake_inventory.lua

SwissalpS:
@fluxionary thanks for chiming in with that very helpful link. Nice collection you have there, a real treasure-chest.

SwissalpS:
closing this in favor of #71

PR #71, opened 2023-06-13
Fix autocrafter variant 3

### LINKED ISSUE #61: autocrafter: if there isn't enough room in the output inventory, replacements may be lost

SwissalpS:
This time going the long way of analyzing dst inv and trying to fit items with some logic instead of just blindly adding to each slot.

fixes #61 
replaces #69
replaces #70

Edit: thanks to @fluxionary  and @S-S-X for the valuable input

### DIFF
--- a/autocrafter.lua
+++ b/autocrafter.lua
@@ -68,6 +68,8 @@ local function autocraft(inventory, craft)
 
 	-- success, so apply to actual output inventory
 	inventory:set_list("dst", sandbox_inv:get_list(""))
+	-- clear sandbox, just in case it matters
+	sandbox_inv:set_list("", {})
 	-- consume materials
 	for itemname, number in pairs(consumption) do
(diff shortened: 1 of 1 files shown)


OgelGames:
Hmm, that seems like a lot of code to fix a simple problem... 🤔 I also feel like I've fixed this before...

OgelGames:
There, much simpler: #72

S-S-X:
If doing this I'd say generic library should be pulled in and preferably required pipeworks code reduced to just few lines.
But I do not know any good library, futil thing seems to pull more dependencies (besides itself) and probably at least 95% of it isn't needed here for anything.

SwissalpS:
> There, much simpler: #72

Yep, I hadn't gone that route because max_stack issues, but I must have overlooked get_free_space()

Will test it tonight.

SwissalpS:
closing in favour of #72

PR #72, opened 2023-06-13
Fix autocrafter destroying replacement items

### LINKED ISSUE #61: autocrafter: if there isn't enough room in the output inventory, replacements may be lost

OgelGames:
Simpler version of #69, #70, #71

Works by counting all the output items and replacements, and then comparing that to the space in the `dst` inventory, including partial stacks.

Or to put it a different way, it subtracts the free space from the output items and checks if there is a remainder.

Fixes #61

### DIFF
--- a/autocrafter.lua
+++ b/autocrafter.lua
@@ -35,17 +35,13 @@ end
 
 local function autocraft(inventory, craft)
 	if not craft then return false end
-	local output = craft.output.item
-
-	-- check if we have enough room in dst
-	if not inventory:room_for_item("dst", output) then	return false end
-	local consumption = craft.consumption
-	local inv_index = count_index(inventory:get_list("src"))
 	-- check if we have enough material available
-	for itemname, number in pairs(consumption) do
+	local inv_index = count_index(inventory:get_list("src"))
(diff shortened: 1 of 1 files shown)


S-S-X:
IMO best approach so far.

autocrafter.lua
```diff
@@ -35,16 +35,38 @@ end
 
 local function autocraft(inventory, craft)
 	if not craft then return false end
-	local output_item = craft.output.item
+	local output = craft.output.item
 
 	-- check if we have enough room in dst
-	if not inventory:room_for_item("dst", output_item) then	return false end
+	if not inventory:room_for_item("dst", output) then	return false end
```
SwissalpS, line 41: lines 40 and 41 can be deleted

SwissalpS:
LGTM, passed my manual tests.

fluxionary:
i'll try to review this within the next day or two, but don't wait on me to approve it if it seems right.

fluxionary:
i tested this w/ the firethorn jelly recipe shown in #61, and things seem to work as i expect.

PR #82, opened 2023-08-22
MineClone formspec compat

nonfreegithub:
mineclone formspec list backgrouds can be get like in i3 mod


wsor4035:
i kinda doubt that i3 and mcl line up that well......also ties the two together for some things....

nonfreegithub:
> i kinda doubt that i3 and mcl line up that well......also ties the two together for some things....

Try this and see how it works well, I can't do it better, if you can just send PR

wsor4035 (approved):
didnt test, looks ok

Issue #83, opened 2023-08-29
Digging pipeworks nodes in mineclone, takes too much time


nonfreegithub:
Dig nodes takes a long time in Mineclone with wooden or stone axes, even with diamond

In MineClone the mod nodes can only be digged with axe, if you break them with the hand or with pickaxe (with pick it makes a lot of sense to dig it) the block breaks.

the nodes have in groups: `dig_generic = 1`

I don't know what it is, but I guess it's so you can dig with your hand too.

PR #84, opened 2023-08-29
Mineclone dig times balanced

nonfreegithub:
It fix https://github.com/mt-mods/pipeworks/issues/83

After many tests i think i've improved the dig times

### DIFF
--- a/autocrafter.lua
+++ b/autocrafter.lua
@@ -289,8 +289,8 @@ minetest.register_node("pipeworks:autocrafter", {
 	description = S("Autocrafter"),
 	drawtype = "normal",
 	tiles = {"pipeworks_autocrafter.png"},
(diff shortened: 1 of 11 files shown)


nonfreegithub:
> seems that pr makes the node digable by any tool,

not any tool, i was digable with axe, now also with pickaxe (glass is digable with pick and tubes are glasslike, also machines are digables with pick)

> and pretty low levels,

I dont think so, wooden pick takes more time than iron pick, have sense

> doesnt really seem that balanced at all..... at least from my understanding of mineclones groups

maybe we can change this so it takes more time to dig the tubes, something that seems very tedious to build complex tube systems.
now it takes a practical time thought for the construction of tube systems, in whose construction you remove and put tubes continuously

PR #85, opened 2023-09-03
MineClone smoker and blast furnace support

nonfreegithub:
- [x] blast furnaces
- [x] smokers


BuckarooBanzay (approved):
i would have done it a bit differently code-wise but LGTM, didn't test that ingame though, merging if there aren't any objections.

PR #88, opened 2023-09-16
Fix crash if add_entity fails

savilli:
`minetest.add_entity(...)` can fail and return `nil` if there are too many entities already. In this case, we try to call `get_luaentity` on `nil` and the server crashes.


BuckarooBanzay:
Not crashing sounds good but i guess you have other problems with that much entities...

PR #90, opened 2023-10-20
prevent MCL awards mod crash when player is nil

nonfreegithub:
In MCL when player is nil (maybe when is a fake player) MCL awards crash the server

### DIFF
--- a/mcl_furnaces.lua
+++ b/mcl_furnaces.lua
@@ -89,10 +89,10 @@ override.after_dig_node = function(pos, oldnode, oldmetadata, digger)
 end
 
 override.on_metadata_inventory_take = function(pos, listname, index, stack, player)
-   if listname == "dst" and player and player:get_player_name() then
-      if stack:get_name() == "mcl_core:iron_ingot" then
+   if listname == "dst" then
+      if stack:get_name() == "mcl_core:iron_ingot" and type(player) == "userdata" then
 	 awards.unlock(player:get_player_name(), "mcl:acquireIron")
-      elseif stack:get_name() == "mcl_fishing:fish_cooked" then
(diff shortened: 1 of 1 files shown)


S-S-X:
It might also fail in some other cases with fake player depending on configuration if it attempts to do this (from awards):
```lua
		local player = minetest.get_player_by_name(name)
		local one = player:hud_add({
```
This might fail if it is fake player, not nil, has a name but player who owns node isn't online when unlock is executed by fake player and awards.show_mode is not set to chat or formspec.

nonfreegithub:
> Maybe awards should be skipped for fake player?

yes I think so

> But it depends if awards should or shouldn't work on fake player.

I think awards should not work with fake players

wsor4035:
tbh this should be fixed in `MCL awards`, not here

corarona:
I can confirm this does not happen with mcl using up-to-date awards from upstream ( e.g. on https://codeberg.org/mineclonia/mineclonia/pulls/640 ) 

For any testers, this setup recreates the bug (mcla / mcl2 with pipeworks):

* Make a redstone pulse setup with 2 observers facing each other
* build a furnace with a iteminjector and pipes leading out (pipes might not be necessary)
* connect the pulser with the iteminjector
* put fuel and raw iron in the furnace
* log out, log in with another name
* run /awards clear
* (with your setup loaded) see it crash

wsor4035:
quick gander, seems https://codeberg.org/mineclonia/mineclonia/commit/7e3064ef934173b420c97106692c108d5d4047e6 is what fixed it

wsor4035:
with confirmation of a mcl* person, closing this

PR #92, opened 2023-11-11
Fix item duplication bug with broken tubes.

debagos:
It was possible to drag items from broken sorting-/mese-tubes.


Niklp09 (approved):

SwissalpS:
looks reasonable to me

debagos:
Sorry for the very brief explanation...
Here is how to reproduce the issue (without the patch applied):
1. Place a sorting tube.
2. Fill all available slots in the sorting tube with diamonds (this won't actually take any diamonds from your inventory since a copy is put into the sorting tubes inventory).
3. Flood the tube with anything until the tube bursts/breaks.
4. Now right click the sorting tube again and drag all the diamonds back into your inventory (yes, you actually get them).
5. Profit!

PR #94, opened 2023-11-15
Teleport tube: replace deprecated inline metadata values ...

### LINKED ISSUE #93: Teleport tube: replace deprecated inline metadata values ...
using a dedicated formspec builder function

This also properly escapes the `S` translation.

I noticed that this mod is not in minetest-mods, thus here's my patch as a file instead of me pushing it right to the repo:

[0001-Teleport-tube-replace-deprecated-inline-metadata-val.patch.txt](https://github.com/mt-mods/pipeworks/files/13367813/0001-Teleport-tube-replace-deprecated-inline-metadata-val.patch.txt)

BuckarooBanzay:
using a dedicated formspec builder function

fixes #93 

CC @SmallJoker

### DIFF
--- a/teleport_tube.lua
+++ b/teleport_tube.lua
@@ -1,5 +1,6 @@
 
 local S = minetest.get_translator("pipeworks")
+local F = minetest.formspec_escape
 local filename = minetest.get_worldpath().."/teleport_tubes"  -- Only used for backward-compat
 local storage = minetest.get_mod_storage()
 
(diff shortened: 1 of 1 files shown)


S-S-X:
If it is just for possible deprecation then I'd wait a moment with this. Last time code for proposed future changes were pushed fast, then core devs decided that changes were not good and result was just more useless code. This is anyway +22 LoC.

If it is for something else too and not just possibly planned future changes then probably fine.

PR #102, opened 2024-01-14
Tube repair fixes

OgelGames:
Just a couple of bug fixes:
- Clear `the_tube_was` metadata value after repair, so the wrench can pick up repaired tubes without complaining.
- Restart vacuum tube node timers after repair, so they work after repairing.


wsor4035 (approved):
makes sense, did not test this.

BuckarooBanzay (approved):

PR #103, opened 2024-01-14
Pipe Hole Texture Toggle setting

JamesClarke7283:
Ive noticed some people don't like the texture of the pipe holes/ tube entry, added to the supported nodes.
This setting allows you to disable that texture if you want to.

This way everyone is happy, its only if the server owner wants the texture or not.
By default the texture is enabled (of course).

Let me know what you think.

### DIFF
--- a/compat-chests.lua
+++ b/compat-chests.lua
@@ -6,7 +6,7 @@
 
 -- Pipeworks Specific
 local fs_helpers = pipeworks.fs_helpers
-local tube_entry = "^pipeworks_tube_connection_wooden.png"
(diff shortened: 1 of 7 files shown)


OgelGames:
This can and should be done with a texture pack. Using a texture pack also means that the changes will apply to all the other mods that support pipeworks and use those textures.

wsor4035:
agree with ogelgames here, this is best done as a texturepack. (servers can have texturepacks).

wsor4035:
due to the above two comments, no comments on discord/irc/matrix channel, closing this

Issue #106, opened 2024-02-04
autocrafter creating replacement items out of nothing


programmerjake:
Reproduction:
Create new world with Minetest Game, pipeworks, basic_materials, and resource_crops.
Create an Autocrafter with the recipe:
Inferno Stone&nbsp;&nbsp;&nbsp;&nbsp;Skeleton Key
Put Skeleton Keys in the Autocrafter's input, but no Inferno Stone.
Turn Autocrafter on.
Observe Autocrafter crafting Gold Ingots and Inferno Stones despite no Inferno Stones in input -- a bug.

OgelGames:
It's caused by this commit: https://github.com/mt-mods/pipeworks/commit/360ed9bdde6158f7e4f3c768a5d4bf15dd69d86f (I knew that PR wasn't ready...)

The bug is here, this function doesn't check the crafting method (as the comment states), unfortunately that means it doesn't ignore cooking recipes.

https://github.com/mt-mods/pipeworks/blob/c87522c52625bc9917d2bd4d9708c1656e87b57e/autocrafter.lua#L27-L64

A quick fix would be to add a check for the crafting type, but maybe that code should be rewritten differently to avoid other bugs.

programmerjake:
> The bug is here, this function doesn't check the crafting method (as the comment states), unfortunately that also means it doesn't ignore cooking recipes.

maybe you misunderstand...the Inferno Stone + Skeleton Key recipe is a standard crafting recipe, not a cooking recipe. or did you mean that the autocrafter looks up the recipe again and gets the other recipe somehow?

SwissalpS:
this particular situation doesn't have to do with crafting method nor with groups :D

I agree that autocrafter can still be improved.

please test #108, I only did very basic testing while debugging this exact case.

SwissalpS:
despite not checking the craft method in ```get_matching_craft```, I have still not been able to misuse autocrafter as a furnace, grinder etc. (with patch #108 applied, of course)

does this call filter them out already? https://github.com/mt-mods/pipeworks/blob/c87522c52625bc9917d2bd4d9708c1656e87b57e/autocrafter.lua#L72-L74

```get_matching_craft``` does then inspect all recipes for given ``output``, but ``output`` has already been limited to 'normal' recipe methods.

I'm still looking for recipes that can trigger other bugs like the one mentioned in this issue.

PR #108, opened 2024-02-06
bugfix autocrafter.lua

SwissalpS:
solves #106 but not for the reason that was assumed. It would be good to check also the crafting method that no cooking, grinding etc. is used


programmerjake:
I did a basic test, it does resolve the issue encountered in #106

OgelGames:
Works to fix the bug, that's all we really need.

programmerjake:
can a new version be published to contentdb?

wsor4035:
> can a new version be published to contentdb?

it should be auto created tonight

PR #111, opened 2024-02-28
Tags support improved

OgelGames:
Replacement for #107. The main differences are:
- Tags are stored in the tube entity, not the item's metadata.
- Tags are sanitized better, including removal of `${key}`.
- Multi-tags work as intended.
- Fixed the tag tube recipe to work with other games, and added an alternate recipe.
- A bunch of other small things.


S-S-X:
Note: changes merged to original branch and are now available on original PR #111

PR #113, opened 2024-03-03
switch to using xcompat

wsor4035:
what the title says, same as all the others


OgelGames:
This will benefit from the texture API: https://github.com/mt-mods/pipeworks/blob/7cc0d128df22888218ddb6ca128a1ca784959d13/tubes/embedded_tube.lua#L5-L6

BuckarooBanzay (approved):

Issue #122, opened 2024-04-01
chest override code is behind minetest_game


fluxionary:
e.g. https://github.com/minetest/minetest_game/pull/3046

i'm not sure whether someone should just copy all the relevant stuff into https://github.com/mt-mods/pipeworks/blob/master/compat-chests.lua, or if the integration should be done in a better way.

fluxionary:
wait it seems like it's already in there, someone told me it wasn't. never mind

S-S-X:
* Yeah, just few weeks ago 😄 #119

PR #123, opened 2024-04-05
de-duplicate chest code

a-tour-ist:
@fluxionary already mentioned in #122 that the chest override code is behind mtg. While the specific bug mentioned there got fixed already, other long-fixed bugs from mtg, such as https://github.com/minetest/minetest_game/pull/1884 still appear with the pipeworks version.

The approach of this PR consists of 2 parts:
1) The override table for the chests is nearly always the same, mostly only changes in the tiles. So I made a general function to override any given chest to be compatible with pipeworks
2) The splitstacks switch in the formspec: 
    - mtg: since mtg opened the chest api (https://github.com/minetest/minetest_game/pull/2127), it's no longer needed to duplicate the open chest code, making it much easier to maintain here. (Just dropping https://github.com/minetest/minetest_game/pull/2965 as another example that got missed here...)
    - hades: got a large chest update... Now all chests from hades connect to pipeworks. I also updated (copy-pasted) some formspec-creation code from there, so a texture error I get when opening a chest is now fixed.
    - mcl_barrels: that switch did not exist in the previous compat code, adding it would mean a lot of code duplication from there... 
    I'd rather opt for a chatcommand/tool to control that value. This would also allow to drop a lot of the code for hades.

I tested everything I can imagine, I hope I missed nothing :-)


SwissalpS:
ran this on a pandorabox-mods world. I didn't test the duplication part, mainly focused on optics and tube functions on technic chests and machines. No obvious errors found this way.

BuckarooBanzay:
thanks :+1:

Issue #124, opened 2024-04-28
crash due to missing nil-check


SwissalpS:
```
2024-04-19 20:27:42: ERROR[Main]: ServerError: AsyncErr: Lua: Runtime error from mod '??' in callback on_playerReceiveFields(): /data/world//worldmods/pipeworks/compat-chests.lua:52: attempt to index a nil value
2024-04-19 20:27:42: ERROR[Main]: stack traceback:
2024-04-19 20:27:42: ERROR[Main]: 	/data/world//worldmods/pipeworks/compat-chests.lua:52: in function 'fn'
2024-04-19 20:27:42: ERROR[Main]: 	.../world//worldmods/monitoring/builtin/received_fields.lua:22: in function <.../world//worldmods/monitoring/builtin/received_fields.lua:20>
2024-04-19 20:27:42: ERROR[Main]: 	/usr/local/share/minetest/builtin/common/register.lua:26: in function </usr/local/share/minetest/builtin/common/register.lua:12>
```

https://github.com/mt-mods/pipeworks/blob/1169cff163f9b0131c41a11d146d111ff065b476/compat-chests.lua#L51-L52

possibly player logged off. Fakeplayer usually has a name, so probably not that.

PR #125, opened 2024-04-28
patch for #124

SwissalpS:
I didn't test this, edited on the fly in browser.

### DIFF
--- a/compat-chests.lua
+++ b/compat-chests.lua
@@ -47,10 +47,7 @@ if minetest.get_modpath("default") then
 	-- get the fields from the chest formspec, we can do this bc. newest functions are called first
 	-- https://github.com/minetest/minetest/blob/d4b10db998ebeb689b3d27368e30952a42169d03/doc/lua_api.md?plain=1#L5840
 	minetest.register_on_player_receive_fields(function(player, formname, fields)
-		if formname ~= "default:chest" then
-			return
-		end
-		if fields.quit then
+		if fields.quit or formname ~= "default:chest" then
 			return
 		end
 		local pn = player:get_player_name()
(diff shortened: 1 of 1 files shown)


compat-chests.lua
```diff
@@ -47,21 +47,25 @@ if minetest.get_modpath("default") then
...
-			-- Do NOT return true here, the callback from default still needs to run
-			return false
+		if formname ~= "default:chest" then
+			return
 		end
+		local pn = player:get_player_name()
+		if not pn then
+			return
+		end
+		local pos = default.chest.open_chests[pn].pos
```
S-S-X, line 57:
```suggestion
		local pos = default.chest.open_chests[pn] and default.chest.open_chests[pn].pos
```
or something like that, see lid close function in mtg and assume that these are all async (because of internet in between)

compat-chests.lua
```diff
@@ -152,17 +161,21 @@ elseif minetest.get_modpath("hades_chests") then
...
-				end
-				-- Do NOT return true here, the callback from hades still needs to run (if they add one)
-				return false
+			if formname ~= "hades_chests:chest_locked" then
+				return
 			end
+			local pn = player:get_player_name()
+			if not pn then
+				return
+			end
```
S-S-X, line 170: This is almost sure unnecessary and could be removed.

S-S-X (approved):
Looks good to me, however kind of skipped style changes and didn't test anything.

Newly added player name check could very likely be dropped but doesn't affect my approval.

compat-chests.lua
SwissalpS: is removed in https://github.com/mt-mods/pipeworks/pull/125/commits/7f63b3f6bf384a325c07ce309f760f06502e212a

SwissalpS:
sorry, saw the second instance now, Line 160 something. Fixing...

PR #126, opened 2024-05-15
Fake player and wielder improvements

### LINKED ISSUE #78: Idea: Open up the wielder API to public use
Today I wanted to look into making a punch happen automatically. I went through pipeworks again and couldn't find what I was after. The mod only includes a deployer, node breaker and dispenser. They are all classed as wielders and live in `wielder.lua`. I was surprised though that the API wasn't open for public use.

The wielder API has a dependency on pipeworks but also on its implementation of fake players. So it should probably continue to live there (though a fake player library that is separate would be good). The wielder code makes it easy to make a machine with an inventory that is activated by mesecons and connected to tubes. I was surprised to find out that the constructor, part of technic, actually uses its own implementation, despite hard-depending on pipeworks.

One could make the case that the punching machine should just be added to pipeworks directly. It probably should. But I think the possibilities are greater if the API were made public. That way people could make other machines we haven't conceived of yet, like setting metadata from a template item, setting nodes (creative-mode kind of machine), swapping nodes, and so on.

OgelGames:
This PR combines several fake player and wielder related changes:
- Replace pipeworks' fake player code with `fakelib` (https://github.com/OgelGames/fakelib).
- Refactor wielder code to be simpler and more generic.
- Expose wielder API with `pipeworks.register_wielder(def)`.
- Add basic digilines functionality to wielders (single activation, with option to specify slot).
- Fix node placement sound not being played to the owner of deployers.

Closes #78, closes https://github.com/pandorabox-io/in-game/issues/95, closes https://github.com/pandorabox-io/in-game/issues/96.


wsor4035:
this can not be merged till the mod is on cdb

BuckarooBanzay (approved):

OgelGames:
@BuckarooBanzay could you put this on the pandorabox test server? Almost everything automated on pandorabox depends on these nodes, so I really want to make sure I haven't broken anything :)

BuckarooBanzay:
> @BuckarooBanzay could you put this on the pandorabox test server? Almost everything automated on pandorabox depends on these nodes, so I really want to make sure I haven't broken anything :)

(sorry for the delay) i updated the mods to the latest modset and added your `fakelib` and this PR, the world might not be that up-to-date though

PR #127, opened 2024-06-16
Get rid of random print() calls

Emojigit:
This PR gets rid of `print()` calls by either removing them or changing them into appropriate `minetest.log` calls. This PR is ready for review.

### DIFF
--- a/default_settings.lua
+++ b/default_settings.lua
@@ -31,7 +31,7 @@ local settings = {
 	delete_item_on_clearobject = true,
 	use_real_entities = true,
 	entity_update_interval = 0,
-	lua_tube_print_behavior = "noop",
+	lua_tube_print_behavior = "log",
 }
 
 pipeworks.toggles = {}
--- a/default_settings.lua
+++ b/default_settings.lua
@@ -70,7 +70,6 @@ for name, value in pairs(settings) do
 	elseif setting_type == "number" then
 		pipeworks[name] = tonumber(minetest.settings:get(prefix..name) or value)
 	else
-		local got_setting = minetest.settings:get(prefix..name)
-		pipeworks[name] = got_setting ~= "" and got_setting or value
+		pipeworks[name] = value
 	end
 end
--- a/default_settings.lua
+++ b/default_settings.lua
@@ -31,6 +31,7 @@ local settings = {
 	delete_item_on_clearobject = true,
 	use_real_entities = true,
 	entity_update_interval = 0,
+	lua_tube_print_behavior = "noop",
 }
 
 pipeworks.toggles = {}
@@ -70,6 +71,7 @@ for name, value in pairs(settings) do
 	elseif setting_type == "number" then
 		pipeworks[name] = tonumber(minetest.settings:get(prefix..name) or value)
 	else
-		pipeworks[name] = value
+		local got_setting = minetest.settings:get(prefix..name)
+		pipeworks[name] = got_setting ~= "" and got_setting or value
 	end
 end
--- a/default_settings.lua
+++ b/default_settings.lua
@@ -31,7 +31,6 @@ local settings = {
 	delete_item_on_clearobject = true,
 	use_real_entities = true,
 	entity_update_interval = 0,
-	lua_tube_print_behavior = "log",
 }
 
 pipeworks.toggles = {}
--- a/tubes/lua.lua
+++ b/tubes/lua.lua
@@ -229,7 +229,7 @@ end
 -------------------------
 
 local function safe_print(param)
-	if minetest.settings:get("pipeworks_lua_tube_print_behavior") == "log" then
+	if (minetest.settings:get("pipeworks_lua_tube_print_behavior") or "log") == "log" then
 		local string_meta = getmetatable("")
 		local sandbox = string_meta.__index
 		string_meta.__index = string -- Leave string sandbox temporarily
--- a/tubes/lua.lua
+++ b/tubes/lua.lua
@@ -232,7 +232,7 @@ local function safe_print(param)
 	local string_meta = getmetatable("")
 	local sandbox = string_meta.__index
 	string_meta.__index = string -- Leave string sandbox temporarily
-	print(dump(param))
+	minetest.log("action", "Lua Tube print: " .. dump(param))
 	string_meta.__index = sandbox -- Restore string sandbox
 end
 
@@ -603,7 +603,7 @@ local function save_memory(pos, meta, mem)
 		meta:set_string("lc_memory", memstring)
 		meta:mark_as_private("lc_memory")
 	else
-		print("Error: lua_tube memory overflow. "..memsize_max.." bytes available, "
+		minetest.log("error", "lua_tube memory overflow. "..memsize_max.." bytes available, "
 				..#memstring.." required. Controller overheats.")
 		burn_controller(pos)
 	end
--- a/tubes/lua.lua
+++ b/tubes/lua.lua
@@ -229,7 +229,7 @@ end
 -------------------------
 
 local function safe_print(param)
-	if pipeworks.lua_tube_print_behavior == "log" then
+	if minetest.settings:get("pipeworks_lua_tube_print_behavior") == "log" then
 		local string_meta = getmetatable("")
 		local sandbox = string_meta.__index
 		string_meta.__index = string -- Leave string sandbox temporarily
--- a/tubes/lua.lua
+++ b/tubes/lua.lua
@@ -232,7 +232,7 @@ local function safe_print(param)
 	local string_meta = getmetatable("")
 	local sandbox = string_meta.__index
 	string_meta.__index = string -- Leave string sandbox temporarily
-	minetest.log("action", "Lua Tube print: " .. dump(param))
+	print(dump(param))
 	string_meta.__index = sandbox -- Restore string sandbox
 end
 
--- a/tubes/lua.lua
+++ b/tubes/lua.lua
@@ -228,9 +228,8 @@ end
 -- Parsing and running --
 -------------------------
 
-local safe_print = function() end
-if pipeworks.lua_tube_print_behavior == "log" then
-	safe_print = function(param)
+local function safe_print(param)
+	if pipeworks.lua_tube_print_behavior == "log" then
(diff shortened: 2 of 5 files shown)


OgelGames:
I think the prints in the Lua tube should stay, makes more sense than logging IMO.

The others are fine to be removed. (one of them was me forgetting to remove it :P)

Emojigit:
> I think the prints in the Lua tube should stay, makes more sense than logging IMO.

Done in ba00a49

S-S-X:
> I think the prints in the Lua tube should stay, makes more sense than logging IMO.

Shouldn't really, while print is easy on ~~better platforms~~ Linux it is way harder to utilize on Windows. And I'd guess there's now days a lot more Windows players than before.

What it ultimately should do is exactly same what mesecons does (besides mod name):
```lua
local function safe_print(param)
	if mesecon.setting("luacontroller_print_behavior", "log") == "log" then
		local string_meta = getmetatable("")
		local sandbox = string_meta.__index
		string_meta.__index = string -- Leave string sandbox temporarily
		minetest.log("action", string.format("[mesecons_luacontroller] print(%s)", dump(param)))
		string_meta.__index = sandbox -- Restore string sandbox
	end
end
```

Emojigit:
> What it ultimately should do is exactly same what mesecons does (besides mod name):

Done in 739ddc6 (BTW, perhaps it would be better if you can utilize GitHub's review function to give suggestions)

tubes/lua.lua
```diff
@@ -229,11 +229,13 @@ end
 -------------------------
 
 local function safe_print(param)
-	local string_meta = getmetatable("")
-	local sandbox = string_meta.__index
-	string_meta.__index = string -- Leave string sandbox temporarily
-	print(dump(param))
-	string_meta.__index = sandbox -- Restore string sandbox
+	if mesecon.setting("luacontroller_print_behavior", "log") == "log" then
```
S-S-X, line 232: Let's call this luatube instead :)
```suggestion
	if mesecon.setting("luatube_print_behavior", "log") == "log" then
```

Or this (after actual node base name lua_tube, adjust settingtypes.txt below accordingly):
```suggestion
	if mesecon.setting("lua_tube_print_behavior", "log") == "log" then
```

Would be also good to add this to settingtypes.txt for GUI config.
Mesecons has this (just changed name), could probably just go with it:
```
# Behavior of print() inside a lua tube. By default, this emits a message into actionstream.
# Set it to noop if you wish to disable that behavior.
mesecon.luatube_print_behavior (Behavior of print) enum log log,noop
```

S-S-X: Actual node base name is lua_tube but I'd say calling it luatube in settings is just fine too, however you (or someone else) likes to have it.

S-S-X:
> perhaps it would be better if you can utilize GitHub's review function to give suggestions

I would probably have used suggestion as I usually do but you can't add suggestions without changes and previous commit did revert changes so there were not enough changed lines for suggestion feature.

tubes/lua.lua
Emojigit: Instead of abusing Mesecon's namespace, I added a new setting, `pipeworks_lua_tube_print_behavior`, controlling Lua tube's behavior. To handle such text input, there are changes in `default_settings.lua`.

Done in 1980ae1b719b6677c79cba60c8a39e3094100de3

tubes/lua.lua
```diff
@@ -603,7 +606,7 @@ local function save_memory(pos, meta, mem)
 		meta:set_string("lc_memory", memstring)
 		meta:mark_as_private("lc_memory")
 	else
-		print("Error: lua_tube memory overflow. "..memsize_max.." bytes available, "
+		minetest.log("error", "lua_tube memory overflow. "..memsize_max.." bytes available, "
```
S-S-X, line 609:
```suggestion
		minetest.log("info", "lua_tube memory overflow. "..memsize_max.." bytes available, "
```
This should probably be `info` rather than error, it is part of mod mechanisms instead of real error.

tubes/lua.lua
```diff
@@ -228,12 +228,15 @@ end
...
-	string_meta.__index = sandbox -- Restore string sandbox
+local safe_print = function() end
+if pipeworks.lua_tube_print_behavior == "log" then
+	safe_print = function(param)
+		local string_meta = getmetatable("")
+		local sandbox = string_meta.__index
+		string_meta.__index = string -- Leave string sandbox temporarily
+		minetest.log("action", string.format("[pipeworks.tubes.lua] print(%s)", dump(param)))
+		string_meta.__index = sandbox -- Restore string sandbox
+	end
```
S-S-X, line 239: Should we still keep this configurable run time or do you think having it only load time is better?

Like how mesecons allows run time configuration and if any this would be one of those configs I could imagine someone changing without restarting game.

tubes/lua.lua
TheEt1234: what if.... it was put as an error message so you *knew* it was memory, and not some overheating error

S-S-X: I mean error stream is meant for actual mod or engine errors and unexpected situations instead of logging normal expected and fully handled situations.
For one `--quiet` command line switch silences info stream but keeps printing to error stream.
Like having "error" written on message is fine if you'd want to distinguish certain game events but actual error stream should be left for actual errors.

tubes/lua.lua
Emojigit: Oh, if mesecons allow runtime configuration then we should too. Though Pipeworks cannot modify the configuration table according to the file yet...

Emojigit: Done in 9cce442, though if we want it to be truly adjustable, we must change `default_settings.lua` so that the values can be reloaded.

S-S-X: To make it configurable runtime you can just do direct call to `minetest.settings.get` like this:

```suggestion
local function safe_print(param)
	if minetest.settings:get("pipeworks_lua_tube_print_behavior") == "log" then
		local string_meta = getmetatable("")
		local sandbox = string_meta.__index
		string_meta.__index = string -- Leave string sandbox temporarily
		minetest.log("action", string.format("[pipeworks.tubes.lua] print(%s)", dump(param)))
		string_meta.__index = sandbox -- Restore string sandbox
	end
end
```

Then you can just use `/set` builtin chat command to change configuration run time.

Emojigit: Done in 0023175

tubes/lua.lua
TheEt1234: > I mean error stream is meant for actual mod or engine errors and unexpected situations instead of logging normal expected and fully handled situations. For one `--quiet` command line switch silences info stream but keeps printing to error stream. Like having "error" written on message is fine if you'd want to distinguish certain game events but actual error stream should be left for actual errors.

i meant error message in the luacontroller, sorry for confusion

TheEt1234 (approved):
lgtm

default_settings.lua
```diff
@@ -31,6 +31,7 @@ local settings = {
 	delete_item_on_clearobject = true,
 	use_real_entities = true,
 	entity_update_interval = 0,
+	lua_tube_print_behavior = "log",
```
S-S-X, line 34: Never used because this setting is read directly to allow changing it without restarts

```suggestion
```

  

Alternatively settings table could be upgraded to handle both static and dynamic settings.
This would be done with additional table for dynamic defaults and __index meta method for settings table.
Only problem with this is that there's no dedicated table for settings in pipeworks but instead it all gets directly into main `pipeworks` namespace and because of this I wouldn't really recommend going there on this PR.

tubes/lua.lua
```diff
@@ -229,11 +229,13 @@ end
 -------------------------
 
 local function safe_print(param)
-	local string_meta = getmetatable("")
-	local sandbox = string_meta.__index
-	string_meta.__index = string -- Leave string sandbox temporarily
-	print(dump(param))
-	string_meta.__index = sandbox -- Restore string sandbox
+	if minetest.settings:get("pipeworks_lua_tube_print_behavior") == "log" then
```
S-S-X, line 232: Moving in default value from global settings table
```suggestion
	if (minetest.settings:get("pipeworks_lua_tube_print_behavior") or "log") == "log" then
```

S-S-X (approved):
Looks good to me, did not test it but seems it should work just fine.

In game lua tube print function now also works same as mesecons.

default_settings.lua
```diff
@@ -70,6 +70,7 @@ for name, value in pairs(settings) do
 	elseif setting_type == "number" then
 		pipeworks[name] = tonumber(minetest.settings:get(prefix..name) or value)
 	else
-		pipeworks[name] = value
+		local got_setting = minetest.settings:get(prefix..name)
+		pipeworks[name] = got_setting ~= "" and got_setting or value
```
OgelGames, line 74: This will produce nil settings, because `get()` returns nil if there is no value set, not an empty string. It should be like this:
```suggestion
		pipeworks[name] = minetest.settings:get(prefix..name) or value
```

OgelGames (approved):

default_settings.lua
S-S-X: Didn't catch this one but this was added for lua tube print configuration, however as it is now done similar to mesecons allowing run time configuration this change isn't actually needed at all and could be removed.

I think this main configuration thing could see rework at some point so that at least new settings would go into separate configuration table instead of straight under `pipeworks`. This would then also allow mixing run time and load time configs under same table and same API (similar how `pipeworks.enable_*` etc. currently works).

If doing that then it should anyway be in its own PR and for this one could also just revert this change / this file to its original state:
```suggestion
		pipeworks[name] = value
```

I'd be fine also with changes, suggestion is mainly because after applying other suggestions this change does not actually do anything anymore. (basically reset to master for this file would work as well).

S-S-X: My approval stays no matter how this gets handled / if this gets updated or not, don't really care if few lines of unused code is updated to more correct version of unused code 😄

Emojigit: I have a mod for settings reloading: https://github.com/C-C-Minetest-Server/settings_loader

It is still a work in progress, but is already performing well on my few mods loading basic values like strings and integers.

PR #128, opened 2024-07-03
Add vacuum tube visualizations

OgelGames:
Adds pickup range visualizations to vacuum tubes using `vizlib`.

Thanks @SwissalpS for the suggestion :)


SwissalpS:
LGTM, I like that you enabled vizlib for both types of tubes.

PR #134, opened 2024-08-24
Optimize move_entities_globalstep_part1

Emojigit:
This PR optimizes `move_entities_globalstep_part1` by reusing the same table and skipping the time-costly garbage collection. Other optimizations are also made, such as avoiding creating pointless vectors.

After some simple profiling, I found out that `for key in table do table[key] = nil end` is faster than garbage-collecting the old table.

This PR is ready for review.

### DIFF
--- a/luaentity.lua
+++ b/luaentity.lua
@@ -67,36 +67,41 @@ end
 minetest.register_on_shutdown(write_entities)
 luaentity.entities_index = 0
 
-local function get_blockpos(pos)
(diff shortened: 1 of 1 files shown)


BuckarooBanzay (approved):
LGTM :+1:

wsor4035 (approved):
dont really see the point to moving `get_blockpos` function, but 🤷. anyways looks good

PR #139, opened 2024-09-09
Caused the server to crash

VinAdmin:
Added a check for nil.
photo_2024-09-09_12-03-32

### DIFF
--- a/item_transport.lua
+++ b/item_transport.lua
@@ -71,7 +71,9 @@ minetest.register_globalstep(function(dtime)
 	for _, entity in pairs(luaentity.entities) do
 		if entity.name == "pipeworks:tubed_item" then
 			local h = minetest.hash_node_position(vector.round(entity._pos))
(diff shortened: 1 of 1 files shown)


OgelGames:
This won't change anything, nil values are already cancelled out by the `or 0`, and it's not the value that is the problem anyway, it's the table index.

It looks like `minetest.hash_node_position` is returning NaN (not nil), which means the value passed to that function is the problem.

VinAdmin:
After this fix, everything started working correctly.

SwissalpS:
I agree with OgelGames, it's the value in ``h`` that is causing the error.

The reason it worked for you after this patch must be somewhere else. Possibly ``entity._pos`` was invalid in the run that crashed your game.

Please try and find a reliable way to reproduce the crash situation.

(Side-note: seriously a screenshot of text?)

VinAdmin:
> (Примечание: серьезно, скриншот текста?)

Unfortunately, the log file was not saved.

I hoped that the error was not only mine.

PR #141, opened 2024-10-20
Allow digging injectors without clearing the filter

Emojigit:
Thie PR allows diging injectors without clearing their filter inventory. I can't see a reason why this was nessessary.

### DIFF
--- a/filter-injector.lua
+++ b/filter-injector.lua
@@ -451,11 +451,6 @@ for _, data in ipairs({
 			if not pipeworks.may_configure(pos, player) then return 0 end
 			return count
 		end,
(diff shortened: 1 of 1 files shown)


OgelGames:
Looks like it's some leftover code from before injectors made copies of items.

OgelGames (approved):

S-S-X (approved):

SwissalpS:
I used this as a feature. If there was something set, players could punch with fast digging tools without fear to dig and loose settings.

wsor4035:
https://xkcd.com/1172/

Issue #64, opened 2023-04-25
Digiline Filter Injector does not receive digiline signals from vertical sources


g-whiz:
~~Digiline signals cannot reach Digiline Filter Injectors via Digiline Conducting Tubes connected to the "To" end of an injector. However, Digiline tubes adjacent to the injectors (i.e. next to one of the sides displaying arrows) do conduct their signals to the injectors. This seems like it may be an issue where injectors are not "connectable" to digilines on their ends, however this cannot be tested with normal digilines.~~

Digiline Filter Injectors do not receive digiline signals from conductrs above/below the injector. This occurs regardless of how the injector is oriented, and no matter what block is used to conduct the digiline signal.

This makes it impossible to use vertically oriented digiline tubes to carry a signal to a vertically oriented digiline filter injector.

fluxionary:
basically, if the tube connects, digiline signal is *not* transmitted; if it doesn't connect, digiline signal *is* transmitted?

S-S-X:
Injectors use default rules which are:
```lua
digilines.rules.default =
{{x=0,  y=0,  z=-1},
{x=1,  y=0,  z=0},
{x=-1, y=0,  z=0},
{x=0,  y=0,  z=1},
{x=1,  y=1,  z=0},
{x=1,  y=-1, z=0},
{x=-1, y=1,  z=0},
{x=-1, y=-1, z=0},
{x=0,  y=1,  z=1},
{x=0,  y=-1, z=1},
{x=0,  y=1,  z=-1},
{x=0,  y=-1, z=-1}}
```
Which does not include positions directly above or directly below.

This makes it ignore input and output positions when injector is rotated in a way that output points either upwards or downwards.

IMO there's 2 acceptable solutions and 2 bad solutions:
1. **Acceptable (and simple):** just add up/down positions to current rule and possibly break few existing setups.
2. **Acceptable (better future but bit more complicated):** PR to digilines mod that adds position information for dynamic rules, then make rules follow injector rotation and possibly break few existing setups.
3. **Bad:** require OgelGames/digicontrol overrides and use those to make dynamic rules.
4. **Bad:** add injector proxy nodes that handle rotation, without LBM or with technic screwdriver it wont necessarily break existing setups but adds both node count and (unnecessary) complexity.
5. ~fork digilines~

Desour:
[...]

PR #142, opened 2024-10-20
Fixes Issue #64

Deathwing777:
Uses a simple boolean setting in default_settings.lua to decide which digiline rules to use in common.lua

### DIFF
--- a/common.lua
+++ b/common.lua
@@ -15,7 +15,17 @@ pipeworks.rules_all = {{x=0, y=0, z=1},{x=0, y=0, z=-1},{x=1, y=0, z=0},{x=-1, y
 		{x=0, y=1, z=0}, {x=0, y=-1, z=0}}
 
 pipeworks.mesecons_rules={{x=0,y=0,z=1},{x=0,y=0,z=-1},{x=1,y=0,z=0},{x=-1,y=0,z=0},{x=0,y=1,z=0},{x=0,y=-1,z=0}}
-pipeworks.digilines_rules={{x=0,y=0,z=1},{x=0,y=0,z=-1},{x=1,y=0,z=0},{x=-1,y=0,z=0},{x=0,y=1,z=0},{x=0,y=-1,z=0}}
+
+local digilines_enabled = minetest.get_modpath("digilines") ~= nil
+if digilines_enabled and pipeworks.use_default_digilines_rules then
+	pipeworks.digilines_rules=digilines.rules.default
+else
+	-- These rules break vertical connectivity to deployers, node breakers, dispensers, and digiline filter injectors
+	-- via digiline conducting tubes. Changing them may break some builds on some servers, so the setting was added
+	-- for server admins to be able to revert to the old "broken" behavior as some builds may use it as a "feature".
+	-- See https://github.com/mt-mods/pipeworks/issues/64
+	pipeworks.digilines_rules={{x=0,y=0,z=1},{x=0,y=0,z=-1},{x=1,y=0,z=0},{x=-1,y=0,z=0},{x=0,y=1,z=0},{x=0,y=-1,z=0}}
+end
 
 pipeworks.liquid_texture = minetest.registered_nodes[pipeworks.liquids.water.flowing].tiles[1]
 if type(pipeworks.liquid_texture) == "table" then pipeworks.liquid_texture = pipeworks.liquid_texture.name end
--- a/common.lua
+++ b/common.lua
@@ -18,7 +18,7 @@ pipeworks.mesecons_rules={{x=0,y=0,z=1},{x=0,y=0,z=-1},{x=1,y=0,z=0},{x=-1,y=0,z
 
 local digilines_enabled = minetest.get_modpath("digilines") ~= nil
 if digilines_enabled and pipeworks.use_default_digilines_rules then
-	pipeworks.digilines_rules=digilines.rules.default
+	pipeworks.digilines_rules=digiline.rules.default
 else
 	-- These rules break vertical connectivity to deployers, node breakers, dispensers, and digiline filter injectors
 	-- via digiline conducting tubes. Changing them may break some builds on some servers, so the setting was added
--- a/common.lua
+++ b/common.lua
@@ -17,12 +17,12 @@ pipeworks.rules_all = {{x=0, y=0, z=1},{x=0, y=0, z=-1},{x=1, y=0, z=0},{x=-1, y
 pipeworks.mesecons_rules={{x=0,y=0,z=1},{x=0,y=0,z=-1},{x=1,y=0,z=0},{x=-1,y=0,z=0},{x=0,y=1,z=0},{x=0,y=-1,z=0}}
 
 local digilines_enabled = minetest.get_modpath("digilines") ~= nil
-if digilines_enabled and pipeworks.use_default_digilines_rules then
+if digilines_enabled and pipeworks.enable_vertical_digilines_connectivity then
 	pipeworks.digilines_rules=digiline.rules.default
 else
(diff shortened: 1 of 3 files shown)


Deathwing777:
At first I was planning on exposing it to the formspec, but then I realized that the rules were being applied at a more global level, so changing the behavior of individual node instances wasn't going to be easy. So I opted for a simple server wide configuration option to either use the working digilines rules or the broken pipeworks digiline rules (because builds on the server use the broken behavior as a feature).

common.lua
```diff
@@ -15,7 +15,17 @@ pipeworks.rules_all = {{x=0, y=0, z=1},{x=0, y=0, z=-1},{x=1, y=0, z=0},{x=-1, y
 		{x=0, y=1, z=0}, {x=0, y=-1, z=0}}
 
 pipeworks.mesecons_rules={{x=0,y=0,z=1},{x=0,y=0,z=-1},{x=1,y=0,z=0},{x=-1,y=0,z=0},{x=0,y=1,z=0},{x=0,y=-1,z=0}}
-pipeworks.digilines_rules={{x=0,y=0,z=1},{x=0,y=0,z=-1},{x=1,y=0,z=0},{x=-1,y=0,z=0},{x=0,y=1,z=0},{x=0,y=-1,z=0}}
+
+local digilines_enabled = minetest.get_modpath("digilines") ~= nil
+if digilines_enabled and pipeworks.use_default_digilines_rules then
+	pipeworks.digilines_rules=digilines.rules.default
```
S-S-X, line 21:
```suggestion
	pipeworks.digilines_rules=digiline.rules.default
```
Yes, I know this is legacy version and was already changed in 2017 but it would still be better to go with it unless you'd like to open another PR updating `digiline` to `digilines` first.

Basically for consistency fix all of them or for now go with what is used elsewhere, having both for luacheck isn't acceptable IMO.

Deathwing777: Where else is digiline used instead of digilines? I found digilines.rules.default in the base digilines mod in presetrules.lua 

I am fine with either as long as it works. I have verified that digiline.rules.default works as well.

eshattow:
I cannot make sense of what is written in the code comments because it is written as a matter of opinion and diverts away from the code changes and the outcomes. Please remove your opinion of if something is "broken" or "old broken" or "new broken" ... really does not matter and is confusing for some admin trying to understand what outcome the option will have. Describe the outcomes and let the admin have their own thoughts about it when selecting the option. Reasons for an option remaining a default being that it was like this before, good, but you must be specific about what version or release the changes happened (or at least a calendar date).

common.lua
S-S-X: > Where else is digiline used instead of digilines?

Didn't check where pipeworks is using legacy namespace but I know it is what is currently defined for luacheck, if you're able to change `digiline` to `digilines` here: https://github.com/mt-mods/pipeworks/blob/222a865e17a109980296ac075aeffcfcd976f79f/.luacheckrc#L13 without breaking luacheck then that would be best option but as it is there I believe it is used elsewhere too.

Pipeworks however apparently isn't using current recommended `digilines` namespace, if it would then luacheck would be failing all the time.

Anyway, I did add issue about it so it does not have to be fixed here. It might involve changing multiple files unrelated to this update and better done in separate pull request:
* #143

Deathwing777:
The code comments are directed at developers not server admins. It consists of a rationalization for keeping an old rule around in the codebase, and allowing server admins to revert to that behavior.

There are different comments, more like what you describe, aimed at server admins in the settings_type.txt file.

I linked to the github issue in the code comments you are referring to. It has all of the detailed information about edge cases that might want the broken rules instead of the rules that don't break expected behavior of nodes. The github issue also connects to the pull request and to the (eventual) inclusion in the repository, so I don't really understand why you would want to put version information or timestamp in the code comments.

I was using "break connectivity" from a digiline (hardware) perspective. As in the signal does not conduct because the circuit is broken/open (as opposed to short/closed). I can see how you misinterpreted it using "broken" in the more general sense of "it does not work". In any case, reading the linked github issue would make the context clear.

I would have an equally hard time understanding what you mean by "ignores vertical connections" Either it is connected or it isn't. Are you doing some filtering to get it to ignore just some messages?

So maybe something like: "When using these rules, digiline signals are NOT conducted vertically, despite any rotation to the node. This breaks expected behavior with digiline conducting tubes. However, some servers may have builds that relied on the lack of vertical connection and may wish to revert to this behavior."

wsor4035:
agree with author that comments are for devs, the ones in settings file are for users

anyways, seems a few luacheck issues to resolve

Deathwing777:
Trailing whitespace in a comment is a luacheck issue apparently.

eshattow:
> So maybe something like: "When using these rules, digiline signals are NOT conducted vertically, despite any rotation to the node. This breaks expected behavior with digiline conducting tubes. However, some servers may have builds that relied on the lack of vertical connection and may wish to revert to this behavior."

Yes, please, and thank you.

settingtypes.txt
```diff
@@ -85,6 +85,11 @@ pipeworks_use_real_entities (Use Real Entities) bool true
 #A value 0.2 or above may cause issues with accelerator tubes.
 pipeworks_entity_update_interval (Entity Update Interval) float 0 0 0.8
 
+# Use the default rules from the digilines mod.
+# If enabled pipeworks devices (filter injector, deployer, dispenser, node breaker, 
+# teleport tube, lua tube, etc) will conduct digiline signals vertically.
```
S-S-X, line 90: Could maybe verify what it actually affects if going to list those, as far as I know this doesn't affect lua tube for example. All other things might be affected but could either verify or leave listings out.

If going to keep listing and including vague "etc" then better to list known exceptions like lua tube and others if any.

Deathwing777: I conducted the testing, and modified the comment. It now reads:

# Use the default rules from the digilines mod.
# If enabled the following devices will connect to digiline networks in the vertical direction:
# digiline filter injector, deployer, dispenser, node breaker, autocrafter
# If disabled, the devices will not be able to send or recieve digiline signals from the top
# or bottom faces, regardless of the node rotation. This breaks expected behavior with digiline 
# conducting tubes, so it is recommended to enable this option unless you have specific builds
# that make use of the lack of vertical digiline connectivity and those are more important to you.

OgelGames:
I think it might be better for the setting to be "enable vertical digilines connectivity" or similar, at least for me it's less confusing that way.

Deathwing777:
Are we ready to merge the pull request?

wsor4035 (approved):
seems legit

Deathwing777:
Anyone else have any feedback? Are we ready to merge?

eshattow:
I've made my review inline at https://github.com/mt-mods/pipeworks/commit/0e370406e5049ff774cdde9cffc0afa8da506436

OgelGames:
I don't know why @eshattow's comment was marked as spam, but this is the review comment, which I agree with:
> minor nit, this is fine but could be better said as `pipeworks_enable_digiline_connection_vertical` to fit the existing pipeworks configuration variable style

Also, the description for the setting needs to be reworded to match the name change.

S-S-X:
If changing those then could also maybe change setting description a bit, I feel it is unnecessarily verbose and probably could be more about facts and less about recommendations. Anyway, it does explain what setting does so this description rewording would be rather low prio stuff.

eshattow:
> If changing those then could also maybe change setting description a bit, I feel it is unnecessarily verbose and probably could be more about facts and less about recommendations. Anyway, it does explain what setting does so this description rewording would be rather low prio stuff.

I do mention this in the review. Simply the editorial and opinion is, after the factual description, two more lines of text that are not needed and may be deleted.

No idea why or why not things are marked as spam. GitHub is closed source.

PR #144, opened 2024-12-01
fix crash from bad digiline msg

a-tour-ist:
The autocrafter already does type checks for digiline messages. But one was missing :wink: 

How to test:

```lua
digiline_send("autocrafter", {
    42,
    true,
    {}
})
```


wsor4035 (approved):
makes sense

PR #146, opened 2024-12-17
Fix #145 caused by #142 Make new feature opt-in 

### LINKED ISSUE #145: Vertical digiline tubes no longer work as they did
See: https://github.com/pandorabox-io/in-game/issues/401
The server didn't change any settings and existing builds are affected by recent vertical signal handling being changed.

SwissalpS:
I don't see why the new feature should force all existing servers to add a setting to opt-out when new servers can opt-in.

Edit: closes: #145

### DIFF
--- a/default_settings.lua
+++ b/default_settings.lua
@@ -31,7 +31,7 @@ local settings = {
 	delete_item_on_clearobject = true,
 	use_real_entities = true,
 	entity_update_interval = 0,
-	enable_vertical_digilines_connectivity = true,
+	enable_vertical_digilines_connectivity = false,
 }
 
 pipeworks.toggles = {}
--- a/settingtypes.txt
+++ b/settingtypes.txt
@@ -92,7 +92,7 @@ pipeworks_entity_update_interval (Entity Update Interval) float 0 0 0.8
 # or bottom faces, regardless of the node rotation. This breaks expected behavior with digiline
 # conducting tubes, so it is recommended to enable this option unless you have specific builds
 # that make use of the lack of vertical digiline connectivity and those are more important to you.
-enable_vertical_digilines_connectivity (Use the default rules from the digilines mod) bool true
+enable_vertical_digilines_connectivity (Use the default rules from the digilines mod) bool false
 
 # if set to true, items passing through teleport tubes will log log where they came from and where they went.
 pipeworks_log_teleport_tubes (Log Teleport Tubes) bool false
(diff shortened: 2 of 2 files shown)


S-S-X:
I thought it was supposed to be like this.

There apparently were also few more comments and disregarded stuff starting around https://github.com/mt-mods/pipeworks/pull/142#issuecomment-2453626894.
Could consider those while at it, not sure if it's important. Probably not that important but more about "maybe nice to have" stuff.

S-S-X (approved):
Approving as I do agree that, in most cases, it is better to keep original behavior as a default behavior.

Also voicing my opinion here about setting description, however don't really care enough for any of this to make any difference with approval:

For setting description (from PR #142) I do disagree with recommendations as those are opinion based rather than fact based:
```diff
This breaks expected behavior with digiline conducting tubes,
so it is recommended to enable this option
- unless you have specific builds that make use of the lack of vertical
- digiline connectivity and those are more important to you.
+ if you need direct vertical digiline connectivity for listed devices. Also
+ does not directly affect tubes despite explicitly mentioning those.
```

Which in turn is also kinda useless and maybe following would be better:

```diff
# If disabled, the devices will not be able to send or recieve digiline signals from the top
- # or bottom faces, regardless of the node rotation. This breaks expected behavior with digiline
- # conducting tubes, so it is recommended to enable this option unless you have specific builds
- # that make use of the lack of vertical digiline connectivity and those are more important to you.
+ # or bottom faces, regardless of the node rotation.
```

SwissalpS:
It is confusing that a toggle named "enable vertical" actually disables vertical tubes :p

SwissalpS:
what about this ^ compromise?

S-S-X:
> what about this ^ compromise?

My opinion? Better. However about _"This breaks expected behavior with digiline conducting tubes"_, is this something that could and should be fixed (a bug) or is it a feature that has to be paired with new setting?

I mean if it actually is a bug then just cut/paste this line into a [new issue](https://github.com/mt-mods/pipeworks/issues/new?title=New%20setting%20breaks%20expected%20behavior%20with%20digiline%20conducting%20tubes).

SwissalpS:
I consider it a bug. As I understand the goal of that change was to improve connectivity, but it obviously can't be solved with a general rule. There needs to be more granularity because what good is it that e.g. autocrafter can get signals from below, if the tubes can't deliver it.

SwissalpS:
Merging as the comment in settingtypes.txt and the bugy behaviour is out of scope of this PR.

Issue #62, opened 2023-03-08
digiline & stackwise filter injectors don't process "split" stacks


fluxionary:
i'm unsure whether this is a bug or more of a documentation issue.

if there's a full stack of items split amongst multiple slots in the "input" inventory, a stackwise filter injector will only send 1 stack at a time. given its name, that makes sense.

however, the digiline filter injector, when given input like the following from a luacontroller, will send nothing, despite a full stack being available:
```lua
digiline_send("injector", {
  exmatch = true,
  name = "default:chest",
  count = 99,
})
```

S-S-X:
Also somewhat related issue https://github.com/mt-mods/pipeworks/issues/15 (early return: it does not process even complete stacks if there's partial too small stack before requested stack)

It is a bug and also discussed many times but it seems there's no dupe issue...  there was one for sure back when repo was still on gitlab.

Whole inventory lookup should be refactored and optimizations added, automated injectors do waste significant amount of time looking through inventory.

Fixing this issue with some bubble gum shouldn't be too hard but I think nobody started it because adding more features without refactoring for optimizations would just make it even more complicated and in the end make refactoring even harder task...

Side note: I do remember that there was already a PR or at least a branch + discussion to "fix" this issue by removing "exmatch" option. Voiced my opinion against it back then and I do still think that exmatch option should stay and is useful or even required for some things.

SwissalpS:
related or duplicate of #15

The4codeblocks:
should be closed?

PR #154, opened 2025-04-26
fix #15 and #62 (process split stacks)

### LINKED ISSUE #15: Digiline Filter Ejector doesn't eject n Items x in slot m, if there is also a slot < m,  with < n items x.
If there are two or more slots with the same item name, where the first of these slots has less items than the second one, you can only eject the second stack or parts of it, if you prior to this eject the first stack.
To reproduce this, place a chest, a digiline ejector and a chest in a row, connect the ejector to a lua-controller, put 49 cobble into the first slot of the left chest and 99 cobble into another slot. Name the ejector's channel "eject" and run the LUA-controller with the following code:
```
if (event.type=="program") then
   local m =
   {
      exmatch = true,
      name = "default:cobble",
      count = 50,
   }
   digiline_send("eject",m)
end
```
or alternatively:
```
if (event.type=="program") then
   local m = "default:cobble 50"
   digiline_send("eject",m)
end
```
Edit: note: random slotseq sometimes leads to successful ejecting.

The4codeblocks:
fixes #15 and #62

### DIFF
--- a/filter-injector.lua
+++ b/filter-injector.lua
@@ -313,10 +313,8 @@ local function punch_filter(data, filtpos, filtnode, msg)
 				return a < b
 			end)
 		end
-		core.log(dump(sposes))
 		local taken = 0
 		for _, spos in ipairs(sposes) do
-			core.log(_)
 			local stack = frominv:get_stack(frominvname, spos)
 			local doRemove = stack:get_count()
 			if fromtube.can_remove then
@@ -334,7 +332,6 @@ local function punch_filter(data, filtpos, filtnode, msg)
 					filtmeta:set_int("slotseq_index", nextpos)
 					set_filter_infotext(data, filtmeta)
 				end
-				local item
 				local count
 				if data.stackwise then
 					count = math.min(stack:get_count(), doRemove)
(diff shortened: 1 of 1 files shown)


BuckarooBanzay:
there are some luacheck issues:

```
    filter-injector.lua:316:3: accessing undefined variable core
    filter-injector.lua:319:4: accessing undefined variable core
    filter-injector.lua:337:11: unused variable item
    filter-injector.lua:355:5: setting non-standard global variable item
    filter-injector.lua:356:35: accessing undefined variable stack
    filter-injector.lua:366:5: setting non-standard global variable item
    filter-injector.lua:369:69: accessing undefined variable item
    filter-injector.lua:378:3: accessing undefined variable item
    filter-injector.lua:379:51: accessing undefined variable item
```

i think you can ignore the warnings about `core` but the others might be relevant.

Is there a test-procedure/schematic to verify this? (sorry for being picky)

The4codeblocks:
> Is there a test-procedure/schematic to verify this? (sorry for being picky)

the test I used was having an injector try to extract 10 items where the 1st stack has 5 and the other stack has the rest, >? 5 depends on if I'm also testing exact count

The4codeblocks:
> there are some luacheck issues

I think I fixed them

BuckarooBanzay:
didn't test locally but i take your word for it, besides: the new code looks a bit cleaner than the old one :shrug:

yl-whosit:
This caused digilines:chest to dupe items, but I'm not sure if this can or even should be fixed here?
https://github.com/minetest-mods/digilines/issues/97
The chest is using a hack that depends on `allow_metadata_inventory_take` being called with the same index as the item that will be taken just before it happens...

PR #156, opened 2025-04-30
digiline → digilines

### LINKED ISSUE #143: Change `digiline` to `digilines`
Yes, I know this is legacy version and was already changed in 2017 but it would still be better to go with it unless you'd like to open another PR updating `digiline` to `digilines` first.

Basically for consistency fix all of them or for now go with what is used elsewhere, having both for luacheck isn't acceptable IMO.

_Originally posted by @S-S-X in https://github.com/mt-mods/pipeworks/pull/142#discussion_r1807941475_

The4codeblocks:
fixes #143

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -10,7 +10,7 @@ globals = {
 
 read_globals = {
 	-- mods
-	"default", "mesecon", "digiline",
+	"default", "mesecon", "digilines",
 	"screwdriver", "unified_inventory",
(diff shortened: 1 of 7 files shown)


BuckarooBanzay:
we now have an unrelated luacheck warning here due to the rename in `.luacheckrc`:

```
Checking tubes/lua.lua                            1 warning

    tubes/lua.lua:727:9: accessing undefined variable digiline
```

not really a blocking issue IMO though

PR #167, opened 2025-06-25
Fix handling of split stacks

### LINKED ISSUE #15: Digiline Filter Ejector doesn't eject n Items x in slot m, if there is also a slot < m,  with < n items x.

### LINKED ISSUE #62: digiline & stackwise filter injectors don't process "split" stacks

### LINKED ISSUE #166: #154 causes incorrect stack handeling
after #154, itemstacks can be merged in a way that should not be possible:
- `stack_max` bypass: fill a chest with cobble and pull `default:cobble 1000` with a digiline injector
- different metadata: create different written books and pull `default:book_written 2`

OgelGames:
Fixes/improves injectors' handling of split stacks in inventories.

I have tested this, but given the previous PR had bugs I would like someone else to double check before it is merged.

Fixes #15, fixes #62, fixes #166.

@a-tour-ist @The4codeblocks

### DIFF
--- a/filter-injector.lua
+++ b/filter-injector.lua
@@ -117,8 +117,7 @@ local function punch_filter(data, filtpos, filtnode, msg)
 	}
 
 	-- make sure there's something appropriate to inject the item into
-	local todir = pipeworks.facedir_to_right_dir(filtnode.param2)
-	local topos = vector.add(filtpos, todir)
+	local topos = vector.add(filtpos, dir)
 	local tonode = minetest.get_node(topos)
 	local todef = minetest.registered_nodes[tonode.name]
 
@@ -314,6 +313,8 @@ local function punch_filter(data, filtpos, filtnode, msg)
 				return a < b
 			end)
 		end
+		local available_items = {}
+		local available_count = 0
 		for _, spos in ipairs(sposes) do
 			local stack = frominv:get_stack(frominvname, spos)
 			local doRemove = stack:get_count()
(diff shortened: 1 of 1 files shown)


a-tour-ist:
The setups I used when testing to report #166 work as intended now. And after reading the responsible code, I can't think of a bad setup either.

IMO fixed, many thanks :-)

PR #174, opened 2025-11-15
Fix missing arrows on lua_tube yellow port

Watilin:
The controller tube (`lua_tube`) had missing arrows on the west and east sides of its yellow port. This PR fixes it.
lua_tube_arrows


SwissalpS:
tested, seems to do what the package says.

PR #178, opened 2025-11-29
fix misnamed setting

a-tour-ist:
missing prefix in settingtypes.txt
https://github.com/mt-mods/pipeworks/blob/2cb3d74d360c77a907e1eb99fc1749a11ca83882/default_settings.lua#L70


wsor4035 (approved):
lgtm, thanks

PR #179, opened 2025-11-30
Make tp tubes use mod storage with prefixed keys

### LINKED ISSUE #177: Teleport tubes forbid other uses of mod storage
Currently, [teleport.lua#L106](https://github.com/mt-mods/pipeworks/blob/2cb3d74d360c77a907e1eb99fc1749a11ca83882/tubes/teleport.lua#L106) throws an error when the storage contains any key that isn't recognized as teleport tube stuff. It prevents other uses of mod storage, for example in #175 (instead of file storage).

I suggest we modify teleport functions to prefix their storage keys, and only handle keys having the relevant prefix. I'm willing to do it. I know that a `tptube` API is exposed, but after a quick look at the code, i think it's possible to do the change transparently and stay with the current API version number.

Watilin:
Change every call to `storage:*` to use prefixed keys. Also cleans up previous prefix-less keys upon migrating.
It is worth noting that this might inadvertently erase keys set by other parts of Pipeworks if those keys happen to match the position hash of an existing teleport tube. But I assume that it won't happen, because I trust that nobody will use prefix-less keys in future extensions of Pipeworks.

Fixes #177.

### DIFF
--- a/tubes/teleport.lua
+++ b/tubes/teleport.lua
@@ -2,6 +2,7 @@
 local S = core.get_translator("pipeworks")
 local filename = core.get_worldpath().."/teleport_tubes"  -- Only used for backward-compat
 local storage = core.get_mod_storage()
+local STORAGE_PREFIX = "tptube_"
 
 local enable_logging = core.settings:get_bool("pipeworks_log_teleport_tubes", false)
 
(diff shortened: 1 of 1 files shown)


BuckarooBanzay (approved):
lgtm, didn't test though

Issue #133, opened 2024-08-12
[suggestion] Better autocrafter


TheEt1234:
When automatically crafting something with multiple ingredients, there often is a problem where the supply of one ingredient is way bigger than the other

Like for example with bonemeal:mulch, you often have way way too many leaves going in, filling the autocrafter with leaves, leaving no space for wood

And that is a tricky problem that the player shouldnt deal with imo

What would be my idea to solve this problem at the pipeworks level:

The ability to optionally allocate slots for certain items

When enabling this option, with a recipe like mulch, it will "allocate" 8 slots for leaves, and 1 slot for wood, and the rest being blocked

By "allocate" i mean restricting the slot to only one item type

Also to avoid having to spam the autocrafter with digiline `"single"` commands, it would be cool if we had a `"all"` digiline command that would just make the autocrafter try to craft the most it can

OgelGames:
I'm already working on a rewrite of the autocrafter :) No idea when I'll be able to finish it, but I'm working on it.

TheEt1234:
> I'm already working on a rewrite of the autocrafter :) No idea when I'll be able to finish it, but I'm working on it.

:eyes: What features will your rewrite have

OgelGames:
> What features will your rewrite have

Better recipe autofill, an option to "recycle" replacement items, some way to empty the main inventory (probably a digiline command), and the feature you suggested but automatic based on the recipe ingredients. Also better groups support.

TheEt1234:
nice

SwissalpS:
I disagree with the premise that the autocrafter should manage space of input items.
This is exactly an interesting challenge to solve using other nodes.

SwissalpS:
The idea of being able to flush input inventory has its appeal but it kinda breaks out of how pipeworks has been working. Up to now, only one inventory is accessible for extraction.
Extending pipeworks with such a feature could be useful for other nodes. [technic] compressors and alloy furnaces come to mind. Sometimes an uneaven amount entered and a remainder blocks changing to using the device for some other material.
Again this can easily be solved in-game by monitoring what and how much is sent into the device.
-1 for that feature from me, even if I understand the appeal.

Also keep in mind that user can change the recipe manually or via digiline at any moment. A self regulating autocrafter would be challenged in those moments.

S-S-X:
> Also to avoid having to spam the autocrafter with digiline "single" commands, it would be cool if we had a "all" digiline command that would just make the autocrafter try to craft the most it can

[...]

Issue #80, opened 2023-08-18
autocrafter: replacement with the same item should remain in the main inventory 


fluxionary:
currently, all replacements are placed in the output inventory, requiring extra machinery to inject the item back into the main inventory for repeated crafting. an example is e.g. from farming: crafting flour from wheat and a mortar and pestle. it'd be nice if the mortar and pestle remained in the input inventory after the crafting.

S-S-X:
replacements should go into output inventory, for example to get replacements out.

OgelGames:
Actually it should be both, with a button to switch it.

SwissalpS:
> It should be both, with a button to switch it.

I don't like this idea.
Players will also want a digilines command to switch it.
Plus it would make the code, we only just added, for detecting space for these items, more complicated and less manageable.
As I understand it, this would not only simply switch between targets but also require inspection of which items to send to which inventory.

It's something, that can easily be solved in-game with a sorting tube, that I don't think it's worth the amount of code to maintain.
Besides now there would be one less thing to solve in-game. Autocrafters are already considered, by some, as OP the way they are.

fluxionary:
i omitted a condition in my request. this should *only* apply to replacements which replace the item with itself.

SwissalpS:
> I omitted a very important condition in my request. this should _only_ apply to replacements which replace the item with itself. Items which are consumed or transformed should continue to go to the output inventory.

Thanks for elaborating.
It was part of my reasoning against implementing this issue's suggestion:
> As I understand it, this would not only simply switch between targets but *also require inspection of which items* to send to which inventory.

This adds at least two more levels of complexity to the code. For something that can be solved with a trivial in-game element: sorting tube at injector's exit side.

fluxionary:
how about an argument for consistency w/ the game engine? if an item in a crafting grid is replaced with itself, it stays in the crafting grid. you don't have to keep moving it back from your main inventory into the grid.

FeXoR-o-Illuria:
"it stays in the crafting grid" This is a horrible behavior in the first place IMO because one manually needs to remove that item from the crafting grid to e.g. not lose it on next death.

IMO those "replace with same" recipes are mostly an annoyance and don't add much to the game.
(And even many of those with containers IMO like bottled/canned/... stuff.
Yes, buckets kind-of make sense to me - then again a wooden bucket would also do - and if that would be expended on use players won't be terribly outraged I guess)

[...]

PR #189, opened 2026-02-20
add function to not use up all ingredients (reserve slot for item-type)

IIIullaIII:
Adds a button, to select whether to keep 1 item in the ingredient slot and prevent other ingredients from filling it. Defaults to off, to completely empty the slots when crafting (current behaviour).

### DIFF
--- a/autocrafter.lua
+++ b/autocrafter.lua
@@ -431,11 +431,9 @@ local function update_meta(meta, enabled)
 	local description, name = get_item_info(output_stack)
 	local infotext = enabled and S("'@1' Autocrafter (@2)", description, name)
 				or S("paused '@1' Autocrafter", description)
-
 	meta:set_string("infotext", infotext)
 	return enabled
 end
-
 -- 1st version of the autocrafter had actual items in the crafting grid
 -- the 2nd replaced these with virtual items, dropped the content on update and
 --   set "virtual_items" to string "1"
--- a/autocrafter.lua
+++ b/autocrafter.lua
@@ -207,58 +207,57 @@ local function has_room_for_output(list_output, index_output)
 	return true
 end
 
-local function autocraft(inventory, craft,pos)
+local function autocraft(inventory, craft, pos)
 	if not craft then return false end
 
-	-- check if output and all replacements fit in dst
 	local output = craft.output.item
 	local out_items = count_index(craft.decremented_input)
-	out_items[output:get_name()] =
-			(out_items[output:get_name()] or 0) + output:get_count()
+	out_items[output:get_name()] = (out_items[output:get_name()] or 0) + output:get_count()
 
 	if not has_room_for_output(inventory:get_list("dst"), out_items) then
 		return false
 	end
-
-	-- check if we have enough material available
-	local inv_index = count_index(inventory:get_list("src"))
+	local inv_list = inventory:get_list("src")
+	local inv_index = count_index(inv_list)
 	local consumption = calculate_consumption(inv_index, craft.consumption)
+	
 	if not consumption then
 		return false
 	end
-
 	--Excludes buckets and non-stackable items
+	--LOGIC OF THE ADDED RESERVE
 	local meta = core.get_meta(pos)
 	if meta:get_int("reserve") == 1 then
+		local total_in_inv = 0
+		local total_needed = 0
+		
 		for itemname, number in pairs(consumption) do
 			local def = core.registered_items[itemname]
 			local stack_max = def and def.stack_max or 1
-			--If it is a bucket or a non-stackable iterm , we ignore it and use it normally
+			
+			--If the object is stackable
 			if stack_max > 1 then
-				if (inv_index[itemname] or 0) <= number then 
-					return false 
-				end
+				total_needed = total_needed + number
+				total_in_inv = total_in_inv + (inv_index[itemname] or 0)
 			end
+		end
+
+		--If the total does not exceed the required total by at least 1, we stop.
+		if total_needed > 0 and total_in_inv <= total_needed then
+			return false
 		end
 	end
 
-	-- consume material
+	--Consumption of objects
 	for itemname, number in pairs(consumption) do
-		-- We have to do that since remove_item does not work if count > stack_max
 		for _ = 1, number do
 			inventory:remove_item("src", ItemStack(itemname))
 		end
 	end
 
-	-- craft the result into the dst inventory and add any "replacements" as well
 	inventory:add_item("dst", output)
-	local leftover
 	for i = 1, 9 do
-		leftover = inventory:add_item("dst", craft.decremented_input[i])
-		if leftover and not leftover:is_empty() then
-			core.log("warning", "[pipeworks] autocrafter didn't " ..
-				"calculate output space correctly.")
-		end
+		inventory:add_item("dst", craft.decremented_input[i])
 	end
 	return true
 end
--- a/autocrafter.lua
+++ b/autocrafter.lua
@@ -25,8 +25,8 @@ local function get_item_info(stack)
 	return description, name
 end
 
--- Get best matching recipe for what user has put in crafting grid.
--- This function does not consider crafting method (mix vs craft)
+-- get best matching recipe for what user has put in crafting grid.
+-- this function does not consider crafting method (mix vs craft)
 local function get_matching_craft(output_name, example_recipe)
 	local recipes = core.get_all_craft_recipes(output_name)
 	if not recipes then
@@ -92,7 +92,7 @@ local function get_craft(pos, inventory, hash)
 	return craft
 end
 
--- From a consumption table with groups and an inventory index,
+-- from a consumption table with groups and an inventory index,
 -- build a consumption table without groups
 local function calculate_consumption(inv_index, consumption_with_groups)
 	inv_index = table.copy(inv_index)
@@ -103,8 +103,8 @@ local function calculate_consumption(inv_index, consumption_with_groups)
 	-- table of ingredients defined as one or more groups each
 	local grouped_ingredients = {}
 
-	-- First consume all non-group requirements
-	-- This is done to avoid consuming a non-group item which
+	-- first consume all non-group requirements
+	-- this is done to avoid consuming a non-group item which
 	-- is also in a group
 	for key, count in pairs(consumption_with_groups) do
 		if key:sub(1, 6) == "group:" then
@@ -141,15 +141,15 @@ local function calculate_consumption(inv_index, consumption_with_groups)
 		return found == count_ingredient_groups
 	end
 
-	-- Next, resolve groups using the remaining items in the inventory
+	-- next, resolve groups using the remaining items in the inventory
 	if next(grouped_ingredients) ~= nil then
 		local take
 		for itemname, count in pairs(inv_index) do
 			if count > 0 then
 				-- groupname is the string as defined by recipe.
-				--  e.g. group:dye,color_blue
+				-- e.g. group:dye,color_blue
 				-- groups holds the group names split into a list
-				--  ready to be passed to core.get_item_group()
+				-- ready to be passed to core.get_item_group()
 				for groupname, groups in pairs(grouped_ingredients) do
 					if consumption_with_groups[groupname] > 0
 						and ingredient_groups_match_item(groups, itemname)
@@ -172,7 +172,7 @@ local function calculate_consumption(inv_index, consumption_with_groups)
 		end
 	end
 
-	-- Finally, check everything has been consumed
+	-- finally, check everything has been consumed
 	for key, count in pairs(consumption_with_groups) do
 		if count > 0 then
 			return nil
@@ -224,8 +224,8 @@ local function autocraft(inventory, craft, pos)
 	if not consumption then
 		return false
 	end
-	--Excludes buckets and non-stackable items
-	--LOGIC OF THE ADDED RESERVE
+	-- ixcludes buckets and non-stackable items
+	--  logic of the added reserve
 	local meta = core.get_meta(pos)
 	if meta:get_int("reserve") == 1 then
 		local total_in_inv = 0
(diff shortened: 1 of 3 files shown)


The4codeblocks:
did you mean to include the .it translation

autocrafter.lua
```diff
@@ -207,7 +207,7 @@ local function has_room_for_output(list_output, index_output)
 	return true
 end
 
-local function autocraft(inventory, craft)
+local function autocraft(inventory, craft,pos)
```
SwissalpS: missing space

autocrafter.lua
```diff
@@ -223,9 +223,17 @@ local function autocraft(inventory, craft)
 	-- check if we have enough material available
 	local inv_index = count_index(inventory:get_list("src"))
 	local consumption = calculate_consumption(inv_index, craft.consumption)
+	
 	if not consumption then
 		return false
 	end
+		-- BLOCk (Riserve 1 item)
```
SwissalpS: No need for all-caps, parenthesis and "reserve" not "Riserve"

SwissalpS: ```-- reserve 1 item``` is sufficient

autocrafter.lua
```diff
@@ -263,7 +271,7 @@ local function run_autocrafter(pos, elapsed)
 	end
 
 	for _ = 1, math.floor(elapsed / craft_time) do
-		local continue = autocraft(inventory, craft)
+		local continue = autocraft(inventory, craft,pos)
```
SwissalpS: missing space

autocrafter.lua
```diff
@@ -344,10 +352,17 @@ local function on_output_change(pos, inventory, stack)
...
--- after this
+
 local function update_meta(meta, enabled)
 	local state = enabled and "on" or "off"
 	meta:set_int("enabled", enabled and 1 or 0)
+	
+	-- Reserve State
+	local res_state = meta:get_int("reserve") == 1
+	local res_icon = res_state and "pipeworks_button_on.png" or "pipeworks_button_off.png"
+	local res_status_txt = res_state and S("RESERVE: ON") or S("RESERVE: OFF")
```
SwissalpS: Why do you feel we need all-caps status text?

autocrafter.lua
```diff
@@ -389,31 +408,37 @@ local function update_meta(meta, enabled)
...
-	local description, name = get_item_info(output)
+	local description, name = get_item_info(output_stack)
 	local infotext = enabled and S("'@1' Autocrafter (@2)", description, name)
 				or S("paused '@1' Autocrafter", description)
 
 	meta:set_string("infotext", infotext)
 	return enabled
 end
 
+
```
SwissalpS: What are all these empty lines for? If I'm not mistaken, pipeworks doesn't usually use this kind of formatting. I'd rather have the deleted comments back ;)

autocrafter.lua
```diff
@@ -484,6 +509,8 @@ core.register_node("pipeworks:autocrafter", {
 		inv:set_size("recipe", 3 * 3)
 		inv:set_size("dst", 4 * 3)
 		inv:set_size("output", 1)
+				meta:set_int("reserve", 0)
```
SwissalpS: Please check indentation here.

autocrafter.lua
```diff
@@ -501,10 +530,22 @@ core.register_node("pipeworks:autocrafter", {
 				start_crafter(pos)
 			end
 		end
+		
+		-- Channel Management (if present)
 		if fields.channel then
 			meta:set_string("channel", fields.channel)
 		end
+
+		--new: Gestion Button Reserve
```
SwissalpS: Consider: ```-- Reserve button management```. The keyword "new" is quickly outdated.

autocrafter.lua
```diff
@@ -501,10 +530,22 @@ core.register_node("pipeworks:autocrafter", {
...
+		if fields.btn_reserve then
+			local current_res = meta:get_int("reserve")
+			local new_res = (current_res == 1 and 0 or 1)
+			meta:set_int("reserve", new_res)
+			-- Aggiorna l'interfaccia per mostrare il cambio colore
```
SwissalpS: Please keep comments in [pipeworks] in English, thanks.

SwissalpS:
Does this work with group ingredients? e.g. mulch -> 8 group:leaves and 1 group:wood

SwissalpS:
For the record: I think it would be better to leave ingredient input management to players. Makes for more interesting in-game solutions than hand-holding via the mod.

The4codeblocks:
> Does this work with group ingredients? e.g. mulch -> 8 group:leaves and 1 group:wood

Is that like just leaving one of a group with no regard for which member it is?

anyway, another solution may just be rejecting inputs that go above a certain capacity quota, may just be 2 slots

OgelGames:
This won't help if the recipe uses a non-stackable item (e.g. a bucket of water); in that case the only way to prevent it from filling is to regulate the input, not the consumption.

IIIullaIII:
> did you mean to include the .it translation
I added too it traduction  and 1 button for chice if keep assignment of the ingredients in the autocraft inventory

autocrafter.lua
IIIullaIII: ok sorry for mystacle i have bad english :-P and I didn't look at the comments in the file, I initially wrote them in Italian for personal use to test the mod

autocrafter.lua
IIIullaIII: It's not necessary, it was just for aesthetics

autocrafter.lua
IIIullaIII: I apologize again and yes I deleted my comments in Italian that I used for testing and maybe I deleted a few too many comments too

autocrafter.lua
IIIullaIII: I apologize again and I did not delete all my comments in Italian that I had used for testing

IIIullaIII:
> For the record: I think it would be better to leave ingredient input management to players. Makes for more interesting in-game solutions than hand-holding via the mod.

I made this with the possibility of choosing whether to activate it or not because in automation with input pipes for example a farm with a chainsaw that has to produce bonemeal will find itself receiving the entire inventory of the autocrafter full of leaves and wood as in the minority it does not have space it is only used for the input automations if it is not needed just have it in off so that it will use all the ingredients

IIIullaIII:
> Does this work with group ingredients? e.g. mulch -> 8 group:leaves and 1 group:wood

yes work with group

IIIullaIII:
> This won't help if the recipe uses a non-stackable item (e.g. a bucket of water); in that case the only way to prevent it from filling is to regulate the input, not the consumption.

Yes, this is an error. In fact, non-stackable items are used and only 1 remain in memory, but it becomes useless. For example, if you put 10 water buckets, only 1 remains and will not be used. Generally, in recipes with buckets, a maximum of 4 are used, which, once used, are sent back to the autocraft after being filled. Instead of using this system, I would opt to remove the non-stackable items from the reserve system.

The4codeblocks:
> > This won't help if the recipe uses a non-stackable item (e.g. a bucket of water); in that case the only way to prevent it from filling is to regulate the input, not the consumption.
> 
> Yes, this is an error. In fact, non-stackable items are used and only remain in memory, but it becomes useless. For example, if you put 10 water buckets, only 1 remains and will not be used. Generally, in recipes with buckets, a maximum of 4 are used, which, once used, are sent back to the autocraft after being filled. Instead of using this system, I would opt to remove the non-stackable items from the reserve system.

what they're saying is that you can run out of space for nonstackables because other items fill the space up

IIIullaIII:
> > > This won't help if the recipe uses a non-stackable item (e.g. a bucket of water); in that case the only way to prevent it from filling is to regulate the input, not the consumption.
> > 
> > 
> > Yes, this is an error. In fact, non-stackable items are used and only remain in memory, but it becomes useless. For example, if you put 10 water buckets, only 1 remains and will not be used. Generally, in recipes with buckets, a maximum of 4 are used, which, once used, are sent back to the autocraft after being filled. Instead of using this system, I would opt to remove the non-stackable items from the reserve system.
> 
> wat they're saying that you can run out of space for nonstackables because other items fill the space up

No, it's just that it's useless to keep a slot occupied if the item isn't stackable, it would only occupy that space without being able to use it, so I removed all the non-stackable ones from the reserve system.

SwissalpS:
This will still not work as hoped. Excluding non-stackables still allows inventory to be blocked by them.

Allowing only one per group will cause items to be rejected that are essential.
E.g. mulch:  two aspen trunks arrive and 10 aspen leaves. 1 mulch is crafted leaving one aspen trunk and two aspen leaves. Now any pine needles and trunks are rejected, only aspen are allowed in.

I'm still convinced this kind of management should be left out of mod code and done by players in-game.
There they can adjust the filter to exactly what they need and not have to find out how exactly the mod
allows and rejects which items in which cases.

About style:
In [pipeworks] we are doing an effort to have **new** comments follow this structure:
```-- First capitalized```
Not ```--No space before first word``` and certainly not ```--ALL CAPS COMMENT```
Spaces after commas and arround operators.

autocrafter.lua
IIIullaIII: fixed

autocrafter.lua
IIIullaIII: fixed

IIIullaIII:
> > > This won't help if the recipe uses a non-stackable item (e.g. a bucket of water); in that case the only way to prevent it from filling is to regulate the input, not the consumption.
> > 
> > 
> > Yes, this is an error. In fact, non-stackable items are used and only remain in memory, but it becomes useless. For example, if you put 10 water buckets, only 1 remains and will not be used. Generally, in recipes with buckets, a maximum of 4 are used, which, once used, are sent back to the autocraft after being filled. Instead of using this system, I would opt to remove the non-stackable items from the reserve system.
> 
> wat they're saying that you can run out of space for nonstackables because other items fill the space up

No I'm saying that they are not stackable like water buckets, they are consumed even if in a craft you never use more than 3 or 4 buckets at a time to make a filling rotation but I removed the non-stackable ones in general especially because it would be useless to keep an unusable full slot I wanted to create an additional window that only assigns the slot without keeping the item, similar to that of sign bot but I noticed that even there the stackable items give problems, so no retention of irtem buchet as the crafting is slow and there are no fast automations such as creating chocolate biscuits the autocrafter fills with wheat and the cocoa does not find space as it is much less and matures in more time, or in the bonemeal farm with leaves and trunks if you cut two trees the autocrafter will be full of leaves and the trunks do not find space and many other recipes require a placeholder, if I made it it is because many people asked me for it also on my server for which I adopted another system, an intelligent distributor among the autocrafter and the injector

IIIullaIII:
> This will still not work as hoped. Excluding non-stackables still allows inventory to be blocked by them.
> 
> Allowing only one per group will cause items to be rejected that are essential. E.g. mulch: two aspen trunks arrive and 10 aspen leaves. 1 mulch is crafted leaving one aspen trunk and two aspen leaves. Now any pine needles and trunks are rejected, only aspen are allowed in.
> 
> I'm still convinced this kind of management should be left out of mod code and done by players in-game. There they can adjust the filter to exactly what they need and not have to find out how exactly the mod allows and rejects which items in which cases.
> 
> About style: In [pipeworks] we are doing an effort to have **new** comments follow this structure: `-- First capitalized` Not `--No space before first word` and certainly not `--ALL CAPS COMMENT` Spaces after commas and arround operators.
1)
''This will still not work as hoped. Excluding non-stackables still allows inventory to be blocked by them.'' 
This is a problem that already exists in the mod, I have excluded it because it makes no sense to keep a slot full of only one element, it would mean taking away a space, and it is difficult to farm with non-stackable objects with the exception of water buckets which must be reused empty and then filled by a nodebreaker, a slow but continuous process.
2)
'' Allowing only one per group will cause items to be rejected that are essential. E.g. mulch: two aspen trunks arrive and 10 aspen leaves. 1 mulch is crafted leaving one aspen trunk and two aspen leaves. Now any pine needles and trunks are rejected, only aspen are allowed in.''
Haven't you tried the mod? Regardless of the name, it uses groups, so it will accept all types of trunks and leaves, but in that specified slot there will always be room for the preferred tree in the farm. Generally, 20 types of trees are not used, but 4 or 5. The excess of leaves in some trees is already an existing problem. The autocrafter fills up with leaves and leaves no room for trunks, and I'm talking about farms with Chinasaw. This way, however, one slot or even one per type is guaranteed. The excess is not accepted, as already happens. In fact, the problem with autocrafter is precisely that it doesn't care about the recipe, but can fill the inventory with anything, even if it's not related to the recipe.
This forces you to use a sorter tube. There are recipes that require up to 9 ingredients, and yes, you could use a lua_sorting_tube, but not everyone knows how to use it. Many younger players prefer a simpler and more intuitive system, and this system does not prevent you from learning and improving your knowledge of using lua sorting. That's why there's the button because you can disable it if you don't need it and you're a pro, but enable it if you still want to play without getting stuck and feeling incompetent.
3)
''> I'm still convinced this kind of management should be left out of mod code and done by players in-game. There they can adjust the filter to exactly what they need and not have to find out how exactly the mod allows and rejects which items in which cases.''
Instead, given the problem described in point 2, I would add a further modification, that is, that the autocrafter only accepts ingredients and groups dedicated to the recipe and not everything that flows in the pipes. In this way, you would already have an autosorter and the possibility of choosing which element to actually keep and which to reject. If you want, I can implement it.
4)
 About style: In [pipeworks] we are doing an effort to have **new** comments follow this structure: `-- First capitalized` Not `--No space before first word` and certainly not `--ALL CAPS COMMENT` Spaces after commas and arround operators.

Yes you are right, I thought I had removed them all, I will check again immediately and I apologize for the oversight

autocrafter.lua
IIIullaIII: fixed

IIIullaIII:
> Does this work with group ingredients? e.g. mulch -> 8 group:leaves and 1 group:wood

yes

SwissalpS:
> Instead, given the problem described in point 2, I would add a further modification, that is, that the autocrafter only accepts ingredients and groups dedicated to the recipe and not everything that flows in the pipes. In this way, you would already have an autosorter and the possibility of choosing which element to actually keep and which to reject. If you want, I can implement it.

No need. All these 'problems' can be managed well in-game to suit whatever situation, tailored to what the player wants. Using luatubes, detector tubes and only injecting what is needed into the autocrafter.

If your players need a 'duplo' version, maybe consider making predefined autocrafters that e.g. only craft mulch another that only crafts canned tomatoes etc.

SwissalpS:
See discussions about similar ideas:
https://github.com/mt-mods/pipeworks/issues/133
#80

IIIullaIII:
> See discussions about similar ideas: #133 #80

Many players want this idea but rightly the final choice is yours

PR #206, opened 2026-06-24
Autocrafter rewrite

### LINKED ISSUE #63: [Autocrafter] Needs a list of non-autocraftable items
There should be some way for mods to mark items as not being able to be crafted by an autocrafter, ideally as a group. This would prevent the autocrafter being able to craft anything in the game, bypassing `minetest.register_on_craft()` and `minetest.register_craft_predict()`.

I suggest the group `do_not_autocraft` as a starting point. If the output item has this group, don't accept the recipe in the crafting grid.

There's currently a [patch thread](https://lists.sr.ht/~gpcf/advtrains-devel/patches/30643) for AdvTrains that would provide survival craft recipes for LuaATC items, items that would require certain privs to craft. Currently the autocrafter makes the priv system (and any other restriction systems) irrelevant as it bypasses all the craft predictions and skips straight to the final output.

### LINKED ISSUE #67: autocrafter does not invoke on_craft callback, resulting in invalid results
e.g. the recipe for crafting a written book + an empty book to create a copy of the empty book results in empty *written* books. also, crafting moretrees coconut milk (coconut + axe + drinking cup) eats the axe for similar reasons. 

see [this approach i took for the unified_inventory_plus "craft all" button](https://github.com/mt-mods/unified_inventory_plus/blob/master/craft_all.lua) for an outline of how to do this correctly, though it will need to be adapted to pipeworks (e.g. the autocrafter doesn't have a real "crafting grid" inventory).

### LINKED ISSUE #80: autocrafter: replacement with the same item should remain in the main inventory

### LINKED ISSUE #133: [suggestion] Better autocrafter

### LINKED ISSUE #162: Auto Crafter does not craft "plasticbox"
Auto crafter doesn't craft "plasticbox:plasticbox" from "basic_materials:plastic_sheet" with switch set to 'on' (see image linked below). Using the digiline command "single" also results in no output.

Image

**Link to plasticbox mod:**
[https://cheapiesystems.com/git/plasticbox/](url)

### LINKED ISSUE #164: Make autocrafter smarter
Gorlock + Hedgehog + ThinkSome

Can autocrafter ensure there is enough room in inventory for all ingredients from the recipe before accepting something in inventory? That would simplify overflow checks.
Or create autocrafter with bigger inventory? Or let autocrafter be able to stack items like in a drawer, and not at a 99 item limit?
Or can autocrafter pull inventory from neighbouring chests, one chest per ingredient.

### LINKED ISSUE #171: Autocrafter nil check missing
2025-07-27 03:37:06: ERROR[Main]: ServerError: AsyncErr: ServerThread::run Lua: Runtime error from mod 'pipeworks' in callback node_on_timer(): .../worldmods/pipeworks_mc/autocrafter.lua:76: bad argument #2 to 'get_craft_result' (table expected, got nil)
2025-07-27 03:37:06: ERROR[Main]: stack traceback: 2025-07-27 03:37:06: ERROR[Main]: [C]: in function 'get_craft_result'
2025-07-27 03:37:06: ERROR[Main]: .../worldmods/pipeworks_mc/autocrafter.lua:76: in function 'get_craft'
2025-07-27 03:37:06: ERROR[Main]: .../worldmods/pipeworks_mc/autocrafter.lua:257: in function <.../worldmods/pipeworks_mc/autocrafter.lua:254>

### LINKED ISSUE #216: Autocrafter destroys written books
## Issue: critical book recipe problem:

Bug Report: Autocrafter Book Metadata Loss

I tested a crafting recipe under the assumption that it would preserve written book metadata. However, the autocrafter consumed the original written book and produced an empty book. This resulted in complete information loss (title, author, and content).

This issue is critical because it prevents photocopying of books via the autocrafter. While manual crafting of book copies correctly preserves metadata, the automated recipe does not.

# **Impact**:

Breaks consistency between manual and automated crafting.

Prevents duplication of written books for libraries, archives, or gameplay mechanics.

Causes irreversible loss of player‑created content when using the autocrafter.

# **Expected Behavior**:

Autocrafter should duplicate written books with full metadata preserved, identical to manual crafting. 

**Here a video which shows the issue** :

https://github.com/user-attachments/assets/fcd4a1a7-8783-4736-9f72-49d4f79ddfac

### LINKED ISSUE #479: Coconut milk/raw coconut destroys axe in autocrafter
Crafting on inventory-grid is fine, but when crafting in autocrafter, the axe goes to /dev/null

### LINKED ISSUE #872: Unreliable crafters and automation
This once used to work, but no longer?

- Autocrafters are no longer crafting some recipes (from several months ago)

  paper to books...
  image

  purple dye and wool to purple wool...
  image

- Water breakers no longer ejecting filled buckets 
  image

  (It used to... And the obvious workaround here is to always have empty buckets in the breaker slot, but eventually it may break by mischance)

OgelGames:
It's been a while since I last did a big rewrite, and like the others, this has been a long time coming.

Here's the big list of changes and new features (most of them at least), in no particular order:

- Crafting now uses actual items from the inventory and calls `on_craft` callbacks, this does reduce performance slightly but fixes many crafting issues.
- Setting a recipe by the output item now works correctly, and tries to select the best recipe available.
- Added an option to keep non-consumables (items replaced with themselves) or all items (for digilines-controlled crafting), with a button to switch.
- There are now more checks for what items will be accepted from tubes, and space will be reserved for missing ingredients (and outputs when keeping them).
- Groups are handled better, and are shown in the recipe grid with "G" on the item.
- Recipe and output lists can now have items shift-clicked out.
- Catch-up is now limited to reduce lag spikes when an autocrafter has been unloaded for a long time.
- Digilines-initiated crafting now uses a queue instead of crafting instantly, with a message sent back when crafting is finished.
- Digilines support has been expanded with more features (including setting the recipe with the output item), and now supports the `command = "action"` message format.
- The crafting time (and the new catch-up limit) can now be adjusted with a setting.
- The autocrafter upgrade function has been removed and replaced with a simple formspec update.

TODO:
- [x] Update formspec code after #209 is merged.
- [x] Lots of testing, preferably on the Pandorabox test server @BuckarooBanzay.
- [x] ~~Update translations with new strings and remove unused ones.~~ To be done in a later PR.
- [x] Implement #213 (#217)

Closes #63, fixes #67, closes #80, closes #133, fixes #162, closes #171, closes #189, fixes #216, fixes pandorabox-io/pandorabox.io#479 (finally), fixes pandorabox-io/pandorabox.io#872, closes pandorabox-io/in-game#164.
I think this might be a record for the most issues closed by one PR :eyes:

Pings for the people who might be interested: @SwissalpS @S-S-X @fluxionary @MCLV-pandorabox


SwissalpS:
I wish the new output redirection's default was what it was historically: output all items.

OgelGames:
I chose to make keeping the replacements the default because that's what happens when you manually craft something, so it's the more intuitive behavior.

Also, only unchanged replacements are kept, so items that are replaced with different items (empty buckets, bottles, etc) are still sent to the output.

SwissalpS:
Some observations:
- manually setting recipe to e.g. 1 trunk and 8 leaves now no longer allows any `group:tree` and `group:leaves`. Only the exact same items as in recipe are used :(
- using digiline setting recipe to `group:tree` and 8 `group:leaves` works however :D
- using digiline to set ``{ { "dye:blue",  "wool:white" } }`` is ignored :(
- using digiline to set ``{ { "dye:blue",  "group:wool" } }`` is also ignored :(

OgelGames:
> * manually setting recipe to e.g. 1 trunk and 8 leaves now no longer allows any `group:tree` and `group:leaves`. Only the exact same items as in recipe are used :(

Yes, this is how it should be, the behavior before was wrong IMO. Also, recipes using groups are preferred when setting the recipe by the output, so it should be possible to manually set those same recipes.

> * using digiline to set `{ { "dye:blue",  "wool:white" } }` is ignored :(
> * using digiline to set `{ { "dye:blue",  "group:wool" } }` is also ignored :(

They were also ignored before, so nothing has changed. Now however, you can use `{command = "set", input = {"dye:blue", "wool:white"}}` to set the recipe without having to specify empty slots.

SwissalpS:
Why doesn't sending the recipe the legacy way just redirect to the new ``command = 'set', input = recipe`` way?

Using the set & input way, even ``{ 'group:dye', 'group:wool' }`` works, thanks.

> Also, recipes using groups are preferred when setting the recipe by the output, so it should be possible to manually set those same recipes.

~That is incorrect as I mentioned in my previous comment.~ oh, the output, but then it just chooses one recipe for the player. e.g. putting mulch on output sets the tree & leaves group recipe, the one based on ``group:seed`` can't be used manually.

OgelGames:
> Why doesn't sending the recipe the legacy way just redirect to the new `command = 'set', input = recipe` way?

It does, but you need to format it correctly, including empty tables: `{ {"dye:blue", "wool:white"}, {}, {} }`

>  but then it just chooses one recipe for the player. e.g. putting mulch on output sets the tree & leaves group recipe, the one based on group:seed can't be used manually

Yeah, items with multiple recipes are a limitation, technic LV battery boxes is another one that comes to mind.

eigen-vectV:
i hope you merge it soon to avoid any further bugs

PR #208, opened 2026-06-29
Updated formspec for Russian localization

VinAdmin:
Hi everyone! Please approve the formspec changes for the injector.

Was

It became

PR #209, opened 2026-07-02
Formspec cleanup and improvements

### LINKED ISSUE #205: Use real coordinates in all formspecs
the old coördinates system is unintuitive and not easy to maintain
ideally forms should be specced with real coördinates in conjunction with extrapolated position values from primitives (no magic numbers)

OgelGames:
This is mostly a code cleanup PR, but with a few formspec changes and improvements.
- Moved `fs_helpers` to a separate file, cleaned-up existing code, and added several new helper functions
- Cleaned-up formspec code using the new helper functions.
- Standardized all formspecs and reduced the amount of text buttons.
- Improved `i3` and `mcl_formspec` compatibility with correct styling.
- Replaced solid color images in sorting tube formspecs with boxes.

The stack-wise injector is the best example of the visual changes, since it includes one of every element. Top row is before, bottom row is after:

formspecs

Fixes #205, closes #208


wsor4035:
we really should make these dynamic instead of hardcoded meta

The4codeblocks:
YEAAAA

The4codeblocks:
> we really should make these dynamic instead of hardcoded meta

explain?

wsor4035:
> > we really should make these dynamic instead of hardcoded meta
> 
> explain?

core.show_formspec instead of encoding it in meta

The4codeblocks:
> > > we really should make these dynamic instead of hardcoded meta
> > 
> > 
> > explain?
> 
> core.show_formspec instead of encoding it in meta

well that's with the compromise of not having the forms autoupdate

OgelGames:
I still don't like the idea of any node formspec being dynamic unless it needs to be.

Good examples are `fancyvend` and `travelnet`, because in both cases the formspecs need to be different depending on who looks at it.

For pipeworks nodes, which only need to update if you change a setting, should not be dynamic in my opinion.

The4codeblocks:
> I still don't like the idea of any node formspec being dynamic unless it needs to be.
> 
> Good examples are `fancyvend` and `travelnet`, because in both cases the formspecs need to be different depending on who looks at it.
> 
> For pipeworks nodes, which only need to update if you change a setting, should not be dynamic in my opinion.

fair

btw, is keeping track of observing players ok in any way?

OgelGames:
> is keeping track of observing players ok in any way?

I'm not sure what you mean?

The4codeblocks:
> > is keeping track of observing players ok in any way?
> 
> I'm not sure what you mean?

keeping track of which players are looking at what forms

SwissalpS:
> > > is keeping track of observing players ok in any way?
> > 
> > 
> > I'm not sure what you mean?
> 
> keeping track of which players are looking at what forms

This seems off-topic.
If you want to observe your players, that is up to you (Disclosing this to your players would be nice).
Use the logs and/or make an observer mod. There is no need to 'infest' other mods with such functionality.
Your mod can catch all 'player-received-fields' callbacks in your observing mod.

The4codeblocks:
> > > > is keeping track of observing players ok in any way?
> > > 
> > > 
> > > I'm not sure what you mean?
> > 
> > keeping track of which players are looking at what forms
> 
> This seems off-topic.
> If you want to observe your players, that is up to you (Disclosing this to your players would be nice).
> Use the logs and/or make an observer mod. There is no need to 'infest' other mods with such functionality.
> Your mod can catch all 'player-received-fields' callbacks in your observing mod.

no I mean the node keeps track of players so that it can send updates when the meta updates

OgelGames:
> the node keeps track of players so that it can send updates when the meta updates

So like a subscription to things that change with that node? I think something like that should be opt-in.

Anyway, SwissalpS is right, this is off-topic. If needed, continue the discussion in discord.

PR #221, opened 2026-08-12
Make autocrafter take items from last stack first

OgelGames:
The autocrafter previously used `InvRef:remove_item()` to take items from it's inventory, which took from the last matching stack first; this PR restores that behavior.

As an added benefit, this change also makes ingredient consumption deterministic, since the code no longer uses `pairs` for iterating.

### DIFF
--- a/autocrafter.lua
+++ b/autocrafter.lua
@@ -1,4 +1,3 @@
-
 local S = core.get_translator("pipeworks")
 local fs_helpers = pipeworks.fs_helpers
 local has_digilines = core.get_modpath("digilines")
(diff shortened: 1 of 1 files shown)


autocrafter.lua
```diff
@@ -239,9 +239,11 @@ end
 
 local function get_ingredients(src, recipe)
 	local items = {}
-	for i, stack in ipairs(src) do
+	-- Work backwards to avoid creating gaps or partial stacks
+	for i=#src, 1, -1 do
```
SwissalpS:
```lua
for i = #src, 1, -1 do
```

autocrafter.lua
```diff
@@ -239,9 +239,11 @@ end
...
+		local stack = src[i]
 		if not stack:is_empty() then
-			items[i] = stack:get_name()
+			items[#items+1] = {stack = stack, name = stack:get_name()}
```
SwissalpS:
```lua
items[#items + 1] = { stack = stack, name = stack:get_name() }
```
(spaces)

